Issue #26, opened 2020-04-14
New Data: Alameda County


elaguerta:
County or Counties: Alameda County
Data source: https://ac-hcsa.maps.arcgis.com/apps/opsdashboard/index.html#/948c67558cff414dbbee1a78fcbab1c9
Output to: 
Update frequency/time: Daily

elaguerta:
REST queries for .json data: 

COVID_counts | https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/COVID_Counts/FeatureServer/0/query?where=0%3D0&outFields=*&f=json
-- | --
COVID_gender_age | https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/COVID_Gender_age/FeatureServer/0/query?where=0%3D0&outFields=*&f=json

COVID_Place_Rates | https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/COVID_Place_Rates/FeatureServer/0/query?where=0%3D0&outFields=*&f=json

Race_eth2 | https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/COVID_Place_Rates/FeatureServer/0/query?where=0%3D0&outFields=*&f=json

PR #34, opened 2020-04-23
#26 Alameda county

elaguerta:
Alameda County data fetcher. To fetch latest data, run the following:
`$ sh install sh`
`$ source env/bin/activate`
`$ python3 data_scrapers/alameda_county.py`

This will log data to stdout.


county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+    """Fetch cases and deaths by age, gender, race, ethnicity
+    Returns a .json by default. If dataframe set to True, returns a pandas
+    DataFrame."""
+    param_list = {'where': "Geography='Alameda County'", 'outFields': '*', 'outSR':4326, 'f':'json'}
+    response = requests.get(demographics, params=param_list)
+    parsed = response.json()
+    fields = parsed['fields']
+    data = parsed['features'][0]['attributes']
+    #make a list of all datapoints with value '<10'
+    # due to some quirk in the json decoding, dashes are stored as the escape sequence '\u2014'
```
Mr0grog, line 126: FWIW, this isn’t a quirk in the decoding — the values just use em dashes (`U+2014`) instead of hyphens (which, typographically, makes sense, since they are being used to delimit a category and a thing, e.g. `Deaths—African_American/Black`). You can see them as em dashes on the open data site, too (e.g. https://data.acgov.org/datasets/AC-HCSA::alameda-county-covid-19-cases-case-rates-and-deaths-by-various-subgroups?selectedAttribute=Deaths_African_American_Black).

Not that this changes anything in practical terms here if what we want at the end of the day is a hyphen instead of an em dash. ¯\\\_(ツ)\_/¯

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+    counts_lt_10 = []
+    for key, val in data.items():
+        key = key.replace('\u2014', '-')
+        if val == '<10':
+            counts_lt_10.append(key)
```
Mr0grog, line 131: Would it be good to validate that other values are integers? e.g:

```py
if val == '<10':
    counts_lt_10.append(key)
else:
    try:
        int(val)
    except ValueError:
        raise ValueError(f'Non-integer value for {key}')
```

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+    if not dataframe:
+        return data, counts_lt_10
+
+    cols = [f['name'] for f in parsed['fields']]
+    dframe = pd.DataFrame(data=data, index=rows, columns=cols)
```
Mr0grog, line 136: Same issue as above — `rows` is never declared, so this raises an error, and I think we could just do `dframe = pd.DataFrame(data)`.

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+        if int(month) < 10:
+            month = '0' + month
+        obj['Date'] = "{}/{}/{}".format(year, month, day)
+    
+    if not dataframe:
+        return features
+    
+    fields = parsed["fields"]
+    cols = [f['name'] for f in parsed['fields']]
+    dframe = pd.DataFrame(data=features, index=rows, columns=cols)
```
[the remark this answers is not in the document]
elaguerta: I don't know if we will even use DataFrames in the pipeline, so I will just strip it out for now. Passing the final dictionaries to DataFrame is working for me. I will add the DataFrame constructors to the comments for get_timeseries() and get_demographics() for reference.

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+    # get age cases
+    out["case_totals"]["age_group"] = { k:v for k,v in demo_totals.items() if 'Age' in k}
+
+    with open('./county_data/alameda_county.json', 'w', encoding = 'utf-8') as f:
+        json.dump(out, f, ensure_ascii= False, indent=4)
+        
+    return json.dumps(out, indent=4) # for printing to console
+
+# Confirmed Cases and Deaths
+def get_timeseries(dataframe = False) -> Dict:
```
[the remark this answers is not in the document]
elaguerta: I've removed the option to return a DataFrame. In the doc-string, I've added a note about creating a DataFrame from the timeseries.

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
elaguerta: Sounds good. I've added an issue for this and will address it in a separate PR.

county_scrapers/alameda_county.py
```diff
@@ -0,0 +1,147 @@
...
+        "meta_from_source": "Notes and disclaimers: The City of Berkeley and Alameda County(minus Berkeley) are separate local health jurisdictions(LHJs). We are showing data for each separately and together. The numbers for the Alameda County LHJ and the Berkeley LHJ come from the state’s communicable disease tracking database, CalREDIE. These data are updated daily, with cases sometimes reassigned to other LHJs and sometimes changed from a suspected to a confirmed case, so counts for a particular date in the past may change as information is updated in CalREDIE. Dates reflect the date created in CalREDIE. Furthermore, we review our data routinely and adjust to ensure its integrity and that it most accurately represents the full picture of COVID-19 cases in our county. The case rates likely reflect more the availability of testing than the actual disease burden. For instance, Hayward and places near Hayward have the highest case rates likely because of the availability of drive-through testing. Berkeley LHJ cases do not include two cases that were passengers of the Diamond Princess cruise.",
```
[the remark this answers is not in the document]
elaguerta: The notes show up after running a long list of js scripts. I was able to pick these up running a webdriver with Selenium. Not sure if this is the most efficient process.

elaguerta:
@Mr0grog @ldtcooper `data_scrapers/alameda_county.py` is ready for review. I'm looking for feedback specifically on the data fetching and the output. I am also looking for suggestions on file organization. I'll open another PR for a command line interface to fetch data that matches the news organization.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+    """Scrape notes and disclaimers from dashboards."""
+    notes = ""
+    driver = webdriver.Firefox()
+    driver.implicitly_wait(30)
+    for url in dashboards:
+        driver.get(url)
+        soup = BeautifulSoup(driver.page_source,'lxml')
+        for p_tag in soup.find_all('p'):
+            if 'Notes' in p_tag.get_text():
+                notes += p_tag.get_text().strip()
```
Mr0grog, line 110: This is only getting the notes from the first page for me. What seems to be happening is that, because everything for these pages is happening client side, the call to load the second URL (`driver.get(url)`), succeeds immediately rather than waiting for the page to finish loading (because no new *page* actually loads), and so `driver.page_source` winds up being the HTML for the loading screen, which obviously has not `<p>` tags.

I know of two straightforward ways to handle that:

1. Open a fresh browser (i.e. call `webdriver.Firefox()`) on each turn of the loop (don’t forget to quit the old one at the end of each turn), or:

2. At the end of each turn of the loop, call `driver.get('about:blank')` to quickly load an empty page.

---

Once the above is working, I think `notes += p_tag_get_text().strip()` may be problematic — it’ll just squish paragraphs from two separate pages together with no space between them.

It might be better to make `notes` a list, and instead of just returning it, return `'\n\n'.join(notes)` (or something along those lines).

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+    return notes
```
Mr0grog, line 111: I think we need to make sure to call `driver.quit()` before returning here. (You could also use a `try...finally` block to quit.)

data_scrapers/alameda_county.py
Mr0grog: You might also want to do a gut-check here by raising an exception if one of the pages *didn’t* have any notes, since (I think?) we expect them both to. If one or both is missing, it’s likely either something changed about the structure of the page being scraped or we got back some kind of error page instead of the actual dashboard.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+        demo_totals["case_totals"]["race_eth"][k] = data[v]
+        # dashes are decoded as the escape sequence '\u2014'
+        demo_totals["death_totals"]["race_eth"][k] = data['Deaths\u2014' + v]
+    # get age cases
+    demo_totals["case_totals"]["age_group"] = {
+        k: v for k, v in data.items() if 'Age' in k}
+
+    #make a list of all datapoints with value '<10'
+    counts_lt_10 = []
+    print(demo_totals)
```
Mr0grog, line 152: Stray debugging print statement.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+    for cat, cat_dict in demo_totals.items(): # cases, deaths
+        for group, group_dict in cat_dict.items(): # dictionaries for age, race/eth
+            for key, val in group_dict.items():
+                if val == '<10':
+                    counts_lt_10.append(cat + ' ' + key)
```
Mr0grog, line 157: It might be helpful to include `group` here, too. Dot notation (e.g. `case_totals.race_eth.African_Amer`) might also be nicer for this use-case than just spaces.

geckodriver.log
```diff
@@ -0,0 +1,112 @@
+1588451853848	mozrunner::runner	INFO	Running command: "/Applications/Firefox.app/Contents/MacOS/firefox-bin" "-marionette" "-foreground" "-no-remote" "-profile" "/var/folders/jc/_d63k96j10l1hc8rrpdfxhn40000gn/T/rust_mozprofileTp33P4"
```
ldtcooper: Is this necessary to commit? If not, we may want to delete this and add `geckodriver.log` to the `.gitignore`

Mr0grog: Ha, somehow I missed that file altogether! Agree this should be in `.gitignore`.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+cases_deaths = 'https://opendata.arcgis.com/datasets/7ea4fd9b8a1040a7b3815f2e0b5f92ba_0/FeatureServer/0/query'
+demographics = 'https://opendata.arcgis.com/datasets/f218564293f9400a8296558e9325f265_0/FeatureServer/0/query'
+cases_meta = 'https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/AC_dates/FeatureServer/0?f=json'
+demographics_meta = 'https://services3.arcgis.com/1iDJcsklY3l3KIjE/arcgis/rest/services/AC_geography/FeatureServer/0?f=json'
+dashboards = ['https://ac-hcsa.maps.arcgis.com/apps/opsdashboard/index.html#/1e0ac4385cbe4cc1bffe2cf7f8e7f0d9',
+              'https://ac-hcsa.maps.arcgis.com/apps/opsdashboard/index.html#/332a092bbc3641bd9ec8373e7c7b5b3d']
+
+# Load data model template into the 'out' dictionary. 'out' will be global to all methods.
+with open('./data_scrapers/_data_model.json') as template:
+    out = json.load(template)
```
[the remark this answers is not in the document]
elaguerta: Agree that using a global is not great. I've moved it into get_county(), and it will return a fresh object by opening the file when called. 
Re: using this as a template vs. a reference, I thought it would be a good way for front-end and other users to see data that is not being collected by the county, since those values would have the default template of "-1". I also thought it would be a good centralized location to change and add labels and data-points if necessary. But maybe it's better to use it as a reference only, or at least grab only the elements of the template for which our sources actually provide data.

data_scrapers/alameda_county.py
elaguerta: It looks nothing now has a value of '<10'. Sadly, maybe that is because these case and death counts have now increased past 10. They are also now using the value `null`. I've edited this logic to include group, use the dot notation, and handle `null` values.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+    """Main method for populating the 'out' dictionary"""
+    # populate dataset headers
+    out["name"] = "Alameda County"
+    out["source_url"] = landing_page
+    out["meta_from_source"] = get_notes()
+
+    # fetch cases metadata, to get the timestamp
+    response = requests.get(cases_meta)
+    response.raise_for_status()
+    cases_header = json.loads(response.text)
```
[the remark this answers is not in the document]
elaguerta: I've changed the json loading to use response.json(). No reason for the change - I'm not familiar with either library and the inconsistencies are probably because I'm referencing examples from different sources.

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,167 @@
...
+import pandas as pd
+from typing import List, Dict
+from datetime import datetime, timedelta, timezone
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+import re
+import os
+
+# Note that we are using numbers for all of Alameda County, including Berkeley
+# Running this scraper requires a Firefox webdriver. The macos Firefox driver, geckodriver, is stored in ./env/bin
```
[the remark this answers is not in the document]
elaguerta: Found a package called [webdrivermanager](https://pypi.org/project/webdrivermanager/) that will do this automatically. I added the following command to 'install.sh':
`webdrivermanager firefox --linkpath ./env/bin;`
Not sure that this is the best way to provide a path to the`bin` directory. When I tested it, by deleting geckodriver and running `sh install.sh`, geckodriver reappeared in my bin folder. However, it resulted in these terminal messages:
`Symlink created: ./env/bin/geckodriver`
`WARNING: Path "./env/bin" is not in the PATH environment variable.`

elaguerta:
@Mr0grog @ldtcooper Thanks so much for your detailed feedback. Ready for another round of review.

data_scrapers/alameda_county.py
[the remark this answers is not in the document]
Mr0grog: > I thought it would be a good way for front-end and other users to see data that is not being collected by the county, since those values would have the default template of "-1". I also thought it would be a good centralized location to change and add labels and data-points if necessary.

Both of these reasons totally make sense! 👍

> maybe it's better to… grab only the elements of the template for which our sources actually provide data.

This might be fine, too. **I think what you’ve got now is totally OK.** If you really wanted to dig in more, something like this might be more flexible:

```py
from functools import lru_cache
from pathlib import Path

# lru_cache will memoize the function. You could also do this manually ¯\_(ツ)_/¯
@lru_cache()
def json_file(filepath):
    with open(filepath) as file:
        return json.load(file)

def county_template():
    # Get path relative to this file, so we don't need to worry
    # about what working directory the script was run from.
    template_path = Path(__file__).parent / '_data_model.json'
    return copy.deepcopy(json_file(template_path))

def get_county() -> Dict:
    # Get a copy of the template to fill in
    out = county_template()

# This version of the function doesn't need the template as input.
# Since the file reading and parsing is cached, it's pretty low
# overhead to ask for a fresh copy (we're just making a copy of data
# already in memory).
def get_demographics() -> (Dict, List):
    out = county_template()['field_you_care_about']
```

But really, what you’ve got now is a-ok. 👌 Let’s not worry *too* much here and get this out the door.

data_scrapers/alameda_county.py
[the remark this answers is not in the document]
Mr0grog: > However, it resulted in these terminal messages:
> `Symlink created: ./env/bin/geckodriver`
> `WARNING: Path "./env/bin" is not in the PATH environment variable.`

You can fix that with this command instead (it’s just complaining because what’s on your PATH is `/absolute/path/to/env/bin` instead of `./env/bin`):

```sh
webdrivermanager firefox --linkpath "$(pwd)/env/bin";
```

BUT I think it *should* also put it there by default if you don’t specify `--linkpath` at all. So maybe you can just drop the option? It seems to work for me, but I also usually set up my virtualenvs differently, so I may be a bad test case.

data_scrapers/alameda_county.py
Mr0grog: Oh, that’s super interesting. Is that for values they just weren’t specifying at all before? Or did `null` replace some other sort of value they used to be using?

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,196 @@
...
+    for cat, cat_dict in demo_totals.items():  # cases, deaths
+        for group, group_dict in cat_dict.items():  # dictionaries for age, race/eth
+            for key, val in group_dict.items():
+                if val == '<10':
+                    counts_lt_10.append(f"{cat}.{group}.{key}")
+                elif val!= None: # if the value wasn't null
+                    try:
+                        int(val)
+                    except ValueError:
+                        raise ValueError(f'Non-integer value for {key}')
```
Mr0grog, line 190: Should `None` be converted to `-1` for these? e.g:

```py
elif val is None:
    group_dict[key] = -1
else:
    try:
        int(val)
    except ValueError:
        raise ValueError(f'Non-integer value for {key}')
```

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
+import requests
+import urllib.request
+from bs4 import BeautifulSoup
+import json
+import pandas as pd
```
Mr0grog, line 6: Since we aren’t actually using it anymore, you can probably drop this import.

requirements.txt
```diff
@@ -8,6 +8,10 @@ idna==2.9
 itsdangerous==1.1.0
 Jinja2==2.11.1
 MarkupSafe==1.1.1
+pandas==1.0.3
```
Mr0grog, line 11: Since we aren’t actually using it anymore, it might make sense to drop this dependency for now.

ldtcooper:
I'll try to review this tonight

data_scrapers/alameda_county.py
Mr0grog: (Since I *think* the whole point of using `-1` instead of `null` was to make sure the values are always numbers and ensure easy parsing for strongly typed tools/languages.)

ldtcooper:
Rob's comments cover what I caught and more. I think with those covered, we'll be good to merge this in

data_scrapers/alameda_county.py
```diff
@@ -0,0 +1,196 @@
...
+    # convert timestamp to datetime object
+    update = datetime.fromtimestamp(timestamp/1000, tz=timezone.utc)
+    out["update_time"] = update.isoformat()
+
+    # get cases, deaths, and demographics data
+    out["series"] = get_timeseries()
+    demo_totals, counts_lt_10 = get_demographics(out)
+    out.update(demo_totals)
+    if counts_lt_10:
+        out["metadata_from_baypd"] = "These datapoints have a value less than 10: " + ", ".join([item for item in counts_lt_10])
```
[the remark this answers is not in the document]
elaguerta: Added an `else:` branch to set the field to the empty string

elaguerta:
@ldtcooper @Mr0grog Just covered these comments. Thanks for your review!

elaguerta:
@Mr0grog @ldtcooper I did a rebase of this branch on master. I think it's ready to be merged. Here's what I checked:
- alameda-county.py works
- files are where they should be in data_models and data_scrapers
- data_models/data_model.json has the most recent changes
I'd like to have a separate README in data_models, so I deleted the age bracket info from the main repo's README.

data_scrapers/san_francisco_county.py
```diff
@@ -3,7 +3,15 @@
...
 
+<<<<<<< HEAD:data_scrapers/san_francisco_county.py
+<<<<<<< HEAD:data_scrapers/san_francisco_county.py
 # API endpoints
+=======
+#EL: get just the AC columns in dte order. Return the latest date as the cumulative case numbers.
+=======
+>>>>>>> Fetch Alameda county cases and deaths into pandas object:county-scrapers/san-francisco-county.py
+# API endpoints 
+>>>>>>> Alameda - fetch age, gender, race:county-scrapers/san-francisco-county.py
```
Mr0grog, line 14: Merge conflict markers need cleanup here.

requirements.txt
```diff
@@ -9,5 +9,12 @@ itsdangerous==1.1.0
...
 requests==2.23.0
+selenium==3.141.0
 urllib3==1.25.8
+webdrivermanager==0.8.0
 Werkzeug==1.0.0
+<<<<<<< HEAD
+pandas==1.0.3
+=======
+
+>>>>>>> e33027196b33ce6c08b2fb8bc0b92d529f7625d0
```
Mr0grog, line 20: Merge conflict markers need cleanup here.

Mr0grog:
My bad — I just realized we didn’t have tests running for PR from forks, so they weren’t running! Can you cleaning up the checks that are failing? (And then let me know if you have a merge button you can click.)

Mr0grog:
Oh, and can you please also fix the PATH message when installing that we discussed? https://github.com/sfbrigade/data-covid19-sfbayarea/pull/34/files#r419833580

I totally missed that after we talked through that it didn’t get changed.

elaguerta:
Updated the path for installing geckodriver, passed checks. I see a clickable merge button.

PR #35, opened 2020-04-23
Add basic news scraper for San Francisco

Mr0grog:
It turns out San Francisco (and all the other bay area counties I've checked so far) don't actually have RSS or other feeds for news, so we need to start scraping these sites if we want to show news across all the counties. This commit adds a really basic scraper that reproduces the current data format used by the site and scrapes San Francisco's coronavirus news page (the site only shows San Francisco right now, so we aren't losing anything here).

You can run it with `sh run_scraper_news.sh` or `python scraper_news.py`.

I didn’t add any tests since there’s no existing structure for them, but let me know if you want me to pull in pytest or something.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,9 @@
+beautifulsoup4==4.9.0
 certifi==2019.11.28
 chardet==3.0.4
 click==7.1.1
 Flask==1.1.1
+html5lib==1.0.1
 idna==2.9
 itsdangerous==1.1.0
 Jinja2==2.11.1
--- a/scraper_news.py
+++ b/scraper_news.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+from bs4 import BeautifulSoup
+import json
+import re
+import requests
+from urllib.parse import urljoin
+
+
+HEADING_PATTERN = re.compile(r'h\d')
+ISO_DATETIME_PATTERN = re.compile(r'^\d{4}-\d\d-\d\d(T|\s)\d\d:\d\d:\d\d(\.\d+)?(Z|\d{4}|\d\d:\d\d)$')
+
+
+def get_base_url(soup, url):
+    base = soup.find('base')
+    if base and base['href']:
+        return urljoin(url, base['href'].strip())
+    else:
+        return url
+
+
+class NewsScraper:
+    START_URL = None
+
+    def scrape(self):
+        # TODO: we may want to iterate through multiple pages in the future
+        response = requests.get(self.START_URL)
+        response.raise_for_status()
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,7 +1,9 @@
+beautifulsoup4==4.9.0
 certifi==2019.11.28
 chardet==3.0.4
 click==7.1.1
 Flask==1.1.1
+html5lib==1.0.1
```
Mr0grog, line 6: `html5lib` is super slow, but we’re not really doing anything high-volume here. Past scraping experience has taught me that `lxml` is fast but not really safe enough for your average web page.

In some other projects, I use [`html5-parser`](https://html5-parser.readthedocs.io/en/latest/), which is a faster version written in C, but installing right requires some fancy pip-specific arguments (so you can no longer use `setup.py install` if you ever want to go down that route). Example: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/a05fd623f5b983db84d4f2370197a61bd4276153/requirements.txt#L8-L9

Let me know if I should change.

scraper_news.py
```diff
@@ -0,0 +1,78 @@
...
+                url = urljoin(base_url, url)
+
+            title = title_link.get_text(strip=True)
+            if not title:
+                raise ValueError(f'No title content found for article {index}')
+
+            date = article.find('time')['datetime']
+            if not ISO_DATETIME_PATTERN.match(date):
+                raise ValueError(f'Article {index} date is not in ISO 8601'
+                                 f'format: "{date}"')
```
Mr0grog, line 59: The current date format is actually `mm/dd/yyy`, but it ultimately gets parsed by a generic date parser that handles lots of formats, so this should be safe (and I think it’s better for the future to use ISO 8601).

See parsing in the Vue code:
- https://github.com/sfbrigade/stop-covid19-sfbayarea/blob/263f8fdd3fb7384d90acdab91b10129aa511d80c/components/WhatsNew.vue#L53-L55
- https://github.com/sfbrigade/stop-covid19-sfbayarea/blob/263f8fdd3fb7384d90acdab91b10129aa511d80c/utils/formatDate.ts#L7-L9

Mr0grog:
For this PR, I’m just trying to match the existing data format from the site (https://github.com/sfbrigade/stop-covid19-sfbayarea/blob/master/data/news.json), but I’m thinking it would be smart to move towards [JSON Feed](https://jsonfeed.org/) in the near future, since it’s a nice, standard format that lots of feed readers support. (We could also add RSS output, too.)

Mr0grog:
Thanks, merging this in.

PR #37, opened 2020-04-25
Requesting Review of SF Common Data Model

kwonangela7:
Hi folks! @elaguerta @kengo-sony @ldtcooper @Mr0grog, @frhino, @collincr, @VincentLa14

Here's a common data model I created for the parsing of just the SF json files (and not the dashboard itself). I then added data points to be parsed that are only present on the dashboard, and I commented those out. I made the distinction since I wasn't sure if the dashboard itself would be scraped.

(Interesting side note - there is data that is in the dashboard that is not in the linked source data i.e. Native American on the race/ethnicity graph)

Some things to mention here:
- Under testing, it made sense to me to keep `pos` and `pct` along with the time series data and not as values associated with a "tabulation" key (like gender/age for cases/deaths) since pos/pct were visualized on a time series graph on the dashboard itself.
- This doesn't account for chhs data (cases: acute v. icu beds, and cases: hospitalized v. note hospitalized); I saw Elaine in her PR mention that we should probably have another table for that. I can work on that after we get feedback on SF/Santa Clara (and before completing the other counties if that's the priority).
- Maybe obvious, but I didn't handle any logic in the template (since as Kengo mentioned in his comment on the related issue, front end might sum stuff up)
- @kengo-sony Re: making sure the model captures all the UI elements
---  "Population per county" isn't something that's explicitly stated on the SF dashboard (don't see it in the JSON files either). I didn't account for external sources just yet in this template. 
- Note that the template includes sample values 

Much of my weekend is open to try to get tasks related to this issue done. Planning on syncing up with @elaguerta tomorrow (hopefully after getting some feedback) to bring our two templates together, and then I can use that to continue to create the data models for the other counties. I figure the common data models for the counties will be very similar, so hopefully I'll be able to finish them faster.

### DIFF
--- a/all_age_brackets_template.json
+++ b/all_age_brackets_template.json
@@ -1,0 +1,92 @@
+{ // Commented out lines are data points only found in the dashboard
+    "County Name": {
+        "data": {
+            "cases": {
+                "age_group_1": { // Age Brackets for SF, Santa Clara, Alameda
+                    "30 or younger": -1,
+                    "31-40": -1,
+                    "41-50": -1,
+                    "51-60": -1,
+                    "61-70": -1,
+                    "71-80": -1,
+                    "81-90" : -1,
+                    "90+": -1,
+                    "Unknown": -1
+                },
+                "age_group_2": { // Marin, Solano
+                    "0-18": -1,
+                    "19-64": -1,
+                    "65+":
+                    
+                },
+                "age_group_3": { // Sonoma, Napa
+                    "0-17": -1,
+                    "18-49": -1,
+                    "50-64": -1,
+                    "64+": -1,
+                    "Unknown": -1
+                },
+                "age_group_4": { // San Mateo
+                    "0 to 19": -1,
+                    "20-29": -1,
+                    "30-39": -1,
+                    "40-49": -1,
+                    "50-59": -1,
+                    "60-69": -1,
+                    "70-79": -1,
+                    "80-89": -1,
+                    "90+": -1
+                    
+                },
+
+                "age_group_5": { // Contra Costa
+                    "0-20": -1,
+                    "21-40": -1,
+                    "41-60": -1,
+                    "61-80": -1,
+                    "81-100": -1,
+                }
+
+            }, 
+            "deaths": { // death by age data only available for SF, Santa Clara, San Mateo. Marin
+                "age_group_1": { // Age Brackets for SF
+                    "Under 60": 1,
+                    "60+": 4
+                },
+                "age_group_2": { // Age Brackets for Santa Clara
+                    "30 or younger": -1,
+                    "31-40": -1,
+                    "41-50": -1,
+                    "51-60": -1,
+                    "61-70": -1,
+                    "71-80": -1,
+                    "81-90" : -1,
+                    "90+": -1,
+                    "Unknown": -1
+                },
+
+                "age_group_3": { // Marin,
+                    "0-18": -1,
+                    "19-64": -1,
+                    "65+": -1
+                    
+                },
+                "age_group_4": { // San Mateo
+                    "0 to 19": -1,
+                    "20-29": -1,
+                    "30-39": -1,
+                    "40-49": -1,
+                    "50-59": -1,
+                    "60-69": -1,
+                    "70-79": -1,
+                    "80-89": -1,
+                    "90+": -1
+                    
+                }
+
+            }
+        }
+
+    }
+    
+}            
--- a/data-model.json
+++ b/data-model.json
@@ -1,7 +1,6 @@
 {   
     "name": "Name1 Name2 County",  
-    "update_date": "yyyy/mm/dd ",
-    "update_time": "hh:mm [am/pm]",
+    "update_time": "yyyy-mm-ddThh:mmz",
     "source_url": "POINT TO LANDING PAGES, IN CASE ENDPOINTS CHANGE", 
     "meta_from_source" : "STORE IMPORTANT NOTES FROM SOURCE HERE",
     "meta_from_baypd": "STORE IMPORTANT NOTES ABOUT OUR METHODS HERE",
@@ -47,23 +46,24 @@
             "unknown": -1
         },
         "age_group": {
-            "30 or younger": -1,
-            "31-40": -1,
-            "41-50": -1,
-            "51-60": -1,
-            "61-70": -1,
-            "71-80": -1,
-            "81" : -1,
+            "30_or_younger": -1,
+            "31_to_40": -1,
+            "41_to_50": -1,
+            "51_to_60": -1,
+            "61_to_70": -1,
+            "71_to_80": -1,
+            "81_to_90" : -1,
+            "90_or_older": -1,
             "Unknown": -1
         },
         "race_eth": {
             "African_Amer": -1,
             "Asian": -1,
-            "Latinx/Hispanic": -1,
+            "Latinx_or_Hispanic": -1,
             "Native_Amer":-1,
             "Multiple_Race":-1,
             "Other": -1,
-            "Pacific Islander":-1,
+            "Pacific_Islander":-1,
             "White":-1,
             "Unknown":-1
         },
@@ -81,31 +81,31 @@
             "unknown": -1
         },
         "age_group": {
-            "20 or younger": -1,
-            "21-30": -1,
-            "31-40": -1,
-            "41-50": -1,
-            "51-60": -1,
-            "61-70": -1,
-            "71-80": -1,
-            "81-90": -1,
-            "90+": -1,
+            "20_or_younger": -1,
+            "21_to_30": -1,
+            "31_to_40": -1,
+            "41_to_50": -1,
+            "51_to_60": -1,
+            "61_to_70": -1,
+            "71_to_80": -1,
+            "81_to_90": -1,
+            "90_or_older": -1,
             "Unknown": -1
         },
         "race_eth": {
             "African_Amer": -1,
             "Asian": -1,
-            "Latinx/Hispanic": -1,
+            "Latinx_or_Hispanic": -1,
             "Native_Amer":-1,
             "Multiple_Race":-1,
             "Other": -1,
-            "Pacific Islander":-1,
+            "Pacific_Islander":-1,
             "White":-1,
             "Unknown":-1
         },
         "underlying_cond": {
             "none":-1,
-            ">1" : -1,
+            "greater_than_1" : -1,
             "unknown": -1
         },
         "transmission_cat": {
@@ -123,25 +123,25 @@
             "unknown": -1
         },
         "age_group": {
-            "20 or younger": -1,
-            "21-30": -1,
-            "31-40": -1,
-            "41-50": -1,
-            "51-60": -1,
-            "61-70": -1,
-            "71-80": -1,
-            "81-90": -1,
-            "90+": -1,
+            "20_or_younger": -1,
+            "21_to_30": -1,
+            "31_to_40": -1,
+            "41_to_50": -1,
+            "51_to_60": -1,
+            "61_to_70": -1,
+            "71_to_80": -1,
+            "81_to_90": -1,
+            "90_or_older": -1,
             "Unknown": -1
         },
         "race_eth": {
             "African_Amer": -1,
             "Asian": -1,
-            "Latinx/Hispanic": -1,
+            "Latinx_or_Hispanic": -1,
             "Native_Amer":-1,
             "Multiple_Race":-1,
             "Other": -1,
-            "Pacific Islander":-1,
+            "Pacific_Islander":-1,
             "White":-1,
             "Unknown":-1
         }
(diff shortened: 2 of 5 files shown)


elaguerta:
@kwonangela7 Thanks for this! I took a pass at combining Santa Clara and SF. See `./data-mode.json`

Questions for @ldtcooper @kengo-sony @Mr0grog @frhino  and other interested parties: 

**Cumulative totals:** Does front-end want to see a cumulative total to date for each day in time series? Sometimes this is provided, in other cases we will have to calculate. In the cases where daily cumulative numbers are provided, sometimes the totals will not add up, and usually there is a comment somewhere explaining this.

**Age bands:** Our models (SF and Santa Clara) did not match up in age bands. Both these counties are going by decades. However, SF starts with "<18, “ “18-30”, while Santa Clara starts with “<20”, “20-30”. As you might expect, neither of these map up neatly to census age bands. I have compromised between SF and Santa Clara by making the first two bands “<30”, “31-40” but we will need to keep an eye on this as we dig into the other counties. 

**Update times:** Most times the data-set for a county seems to be updated on the same day, but timestamps are slightly different. How should we handle multiple times in a county?

**Capitalization:** is there a preference on capitalization of the field names? Dashes or underscores? Seems like underscores are preferred in the source data.

**Transmissions:** SF has a time series by Transmission (new cases by community spread, from contact, or unknown). I think they’re the only county with transmission category x date. SF is only visualizing totals on their dashboard. Do we want to include a series for Transmission, or just totals?

kwonangela7:
Thanks for all the comments Elaine! They were very helpful :) 

Re: age brackets - I screenshotted all the different age groupings here if you want to take a look, but I'll also summarize my findings/thoughts here. Note that SF has a different age grouping (just two buckets) used for deaths. All other counties have the same age buckets for cases and deaths (that is, if they have a deaths by age in the first place).

- Marin and Solano can easily be merged as they share the same age groupings. Napa has similar brackets, but slightly differ by one year when looking at the start or end of a bracket and the start and end of another from Marin/Solano. Sonoma has similar issues; its age bracket is also most closely related to Marin/Solano.

- Alameda's age groupings can be merged with SF/Santa Clara, as it has the the same age buckets are SF, but are missing some of them (they skip the "18-30" bracket, among others.).
- San Mateo is most closely related to SF/Santa Clara/Alameda, but similar to the issue that Napa has with Marin/Solano, SM's brackets differ by 1-2 years when looking at the start or end of a bracket.

- Contra Costa's age groupings are most different from all the other counties.

kwonangela7:
by the way, per Elaine's request I went ahead and created a quick template for all the age groups for all counties for cases and deaths in one json file.

data-model.json
```diff
@@ -0,0 +1,149 @@
+{   
+    "name": "Name1 Name2 County",  
+    "update_date": "yyyy/mm/dd ",
+    "update_time": "hh:mm [am/pm]",
+    "source_url": "POINT TO LANDING PAGES, IN CASE ENDPOINTS CHANGE", 
+    "meta_from_source" : "STORE IMPORTANT NOTES FROM SOURCE HERE",
+    "meta_from_baypd": "STORE IMPORTANT NOTES ABOUT OUR METHODS HERE",
+    "series": {
+        "cases": [
+            { "date": "yyyy/mm/dd", "cases": -1, "cumul_cases": -1},
```
Mr0grog, line 10: Is `cumul_cases` here an example of what you meant by:

> **Cumulative totals:** Does front-end want to see a cumulative total to date for each day in time series? …In the cases where daily cumulative numbers are provided, sometimes the totals will not add up, and usually there is a comment somewhere explaining this.

Either way, that’s super interesting. If summing the numbers is not necessarily correct in all cases, then it seems like we should definitely put the cumulative total in there.

Is the comment usually at the level of the whole dataset, or on a given day? Would it go in `meta_from_source` here?

data-model.json
```diff
@@ -0,0 +1,149 @@
+{   
+    "name": "Name1 Name2 County",  
+    "update_date": "yyyy/mm/dd ",
+    "update_time": "hh:mm [am/pm]",
```
Mr0grog, line 4: FWIW, I think 24 hour time would be preferable here. Basically, treat this as the time component of an W3C-style ISO8601 datetime.

I also might have misunderstood from the Slack conversation, but I was thinking the issue with separating dates and times was just for each item in the timeseries, and for the whole-dataset-update-time, we could just have an actual datetime, e.g. `yyyy-mm-ddThh:mmz`. If I got that wrong, though, that’s ok. I’m not looking to open up a whole new can of worms here. :)

ldtcooper: I agree that 24h time is better here

ldtcooper:
> **Cumulative totals:** Does front-end want to see a cumulative total to date for each day in time series? Sometimes this is provided, in other cases we will have to calculate. In the cases where daily cumulative numbers are provided, sometimes the totals will not add up, and usually there is a comment somewhere explaining this.

> **Age bands:** Our models (SF and Santa Clara) did not match up in age bands. Both these counties are going by decades. However, SF starts with "<18, “ “18-30”, while Santa Clara starts with “<20”, “20-30”. As you might expect, neither of these map up neatly to census age bands. I have compromised between SF and Santa Clara by making the first two bands “<30”, “31-40” but we will need to keep an eye on this as we dig into the other counties.
> 
> **Update times:** Most times the data-set for a county seems to be updated on the same day, but timestamps are slightly different. How should we handle multiple times in a county?
> 
> **Population Totals:** Alameda county provides age, gender, and race/eth breakdowns for the total population, to calculate case rates for each demographic group. I think it is important to somehow normalize numbers by population totals get a sense of any disparate impact on demographic groups. Most counties however, don't provide these totals. For those that do not, should we source these all from a common census estimate? Perhaps the [2018 ACS 1 -year estimate](https://data.census.gov/cedsci/table?q=Santa%20Clara%20County&g=0500000US06085&hidePreview=false&tid=ACSDP1Y2018.DP05&vintage=2018&layer=VT_2018_050_00_PY_D1&cid=DP05_0001E)?
> 
> **Capitalization:** is there a preference on capitalization of the field names? Dashes or underscores? Seems like underscores are preferred in the source data.
> 
> **Transmissions:** SF has a time series by Transmission (new cases by community spread, from contact, or unknown). I think they’re the only county with transmission category x date. SF is only visualizing totals on their dashboard. Do we want to include a series for Transmission, or just totals? @kengo-sony the UX wireframes do not show time series by transmission.
> 
> **Hospitalizations:** Since we are planning to use [the chhs](https://data.chhs.ca.gov/dataset/california-covid-19-hospital-data-and-case-statistics/resource/6cd8d424-dfaa-4bdd-9410-a3d656e1176e?inner_span=True) as a common source for hospital data per county, perhaps we should keep it in a separate table. @kengo-sony the UX wireframes are not currently visualizing hospital data.

ldtcooper:
Whoops, accidentally pressed a hotkey while trying to write a comment

data-model.json
elaguerta: It's at the level of the dataset, so the comment would go in meta_from_source.

data-model.json
elaguerta: Yes, I agree with having a whole-dataset-update-time per county. Would this format: `yyyy-mm-ddThh:mmz`  be a string? Some of the datasets I've seen on arcgis are using timestamps (ms since 1970/1/1).

elaguerta:
**Action Items**
Just want to keep a running list here. I will edit this comment to confirm and summarize the discussion.

1. Add hospitalization data from chhs to data model for each county. Perhaps it needs its own metadata to track source and update times.
2. Add fields for all calculated numbers showing up in the UX wireframe, i.e. case-rates per 1000 people
3. Add separate entries for state and country population and case rates (I will add this as a new data issue)

data-model.json
Mr0grog: I think either is ok, but I generally much prefer [the W3C take on ISO 8601](https://www.w3.org/TR/NOTE-datetime) (i.e. the format you just suggested, @elaguerta). It has 2 big benefits:
1. Somewhat human readable while still being easy to parse and sortable, and
2. No confusion over whether the timestamp is milliseconds (friendly to JS) or seconds (more common for most other libraries and *nix tools) since Unix epoch.

Mr0grog: (It can also carry time zone info, which is super valuable, even when dealing with data that’s always coming from one geographic area, since most servers are configured to use UTC.)

kengoy: I prefer ISO 8601 format as well. ex) "2020-04-26T08:00:00Z"

kengoy:
> @kengo-sony Re: making sure the model captures all the UI elements
> --- "Population per county" isn't something that's explicitly stated on the SF dashboard (don't see it in the JSON files either). I didn't account for external sources just yet in this template.

Thanks for catching this up. Understand. I think we can just use a static population number based on a specific data source like Census 2019 for the calculation.

data-model.json
```diff
@@ -0,0 +1,149 @@
...
+            "Unknown": -1
+        },
+        "race_eth": {
+            "African_Amer": -1,
+            "Asian": -1,
+            "Latinx/Hispanic": -1,
+            "Native_Amer":-1,
+            "Multiple_Race":-1,
+            "Other": -1,
+            "Pacific Islander":-1,
```
elaguerta, line 102: "Pacific Islander" should have an underscore.

data-model.json
kwonangela7: "If summing the numbers is not necessarily correct in all cases, then it seems like we should definitely put the cumulative total in there."  - I think I'm misunderstanding; if summing the numbers is not necessarily correct, wouldn't we not want to put the cumulative total in there? (since the sum isn't correct0

data-model.json
kwonangela7: So the final decision is that we're using `yyyy-mm-ddThh:mmz` correct?

all_age_brackets_template.json
```diff
@@ -0,0 +1,92 @@
...
+                    "90+": -1,
+                    "Unknown": -1
+                },
+
+                "age_group_3": { // Marin,
+                    "0-18": -1,
+                    "19-64": -1,
+                    "65+": -1
+                    
+                },
```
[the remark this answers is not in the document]
kwonangela7: Earlier in the weekend it seemed that we wanted to merge the age brackets when possible, but just read your comment above - "If each county provides different age brackets, the easiest way around this may be to simply use each county's provided age brackets rather than try to normalize them. I'm worried that normalizing them will lose a lot of the granularity that makes the age data interesting. I think that the charting library should be able to handle different brackets. We can make a note of why the brackets differ from county to county.". 

If I'm understanding things correctly, we can just ignore this age brackets json since we won't be merging age brackets. That's fine with me!

data-model.json
Mr0grog: I was reacting to this:

> In the cases where daily cumulative numbers are provided, sometimes the totals will not add up, and usually there is a comment somewhere explaining this.

Which is saying, if I’m reading correctly, that some counties *provide* cumulative totals rather than us having to synthesize them by summing the dailies, and that in some cases a synthesized cumulative total wouldn’t match the provided one anyway because of some quirk of the data or because of some other information not surfaced through the API (and usually a description of the discrepancy would be in the metadata).

So based on that, I was thinking that we should include the cumulative total because, in some cases, the cumulative total comes from the county and we couldn’t reproduce it ourselves, even if we downloaded all the data on the front-end.

Does that make sense?

data-model.json
Mr0grog: Yep, it seems like we’re all largely in agreement on that. 👍

data-model.json
```diff
@@ -0,0 +1,149 @@
+{   
+    "name": "Name1 Name2 County",  
+    "update_time": "yyyy-mm-ddThh:mmz",
+    "source_url": "POINT TO LANDING PAGES, IN CASE ENDPOINTS CHANGE", 
+    "meta_from_source" : "STORE IMPORTANT NOTES FROM SOURCE HERE",
+    "meta_from_baypd": "STORE IMPORTANT NOTES ABOUT OUR METHODS HERE",
+    "series": {
+        "cases": [
+            { "date": "yyyy/mm/dd", "cases": -1, "cumul_cases": -1},
+            { "date": "yyyy/mm/dd", "cases": -1, "cumul_cases": -1 }
```
Mr0grog, line 10: So I totally missed this before, but `yyyy/mm/dd` seems like a super unusual format. Could we use ISO 8601 dates instead? (`yyyy-mm-dd` — just swap the `/` for `-`.)

all_age_brackets_template.json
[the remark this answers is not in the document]
ldtcooper: Yeah, the more I think about this, I think it makes more sense to not merge the age groups

kwonangela7:
@ldtcooper @Mr0grog @elaguerta think I tied up the loose ends in my most recent commit - hopefully I didn't miss anything, and hopefully data_model.json is good to go!

By the way, something I mentioned earlier in the call was that I could look through each county's dashboards to see if there were any other inconsistencies to note (similar to how I noted the age brackets), but after the rest of the conversation that doesn't seem necessary.. and hopefully we're on the same page about that as well. Let me know if not though.

PR #44, opened 2020-04-30
Restructure news scrapers into `news` package

### LINKED ISSUE #41: Create directory structure to organize news scrapers
All the code for scraping news sites currently lives in one file: [`scraper_news.py`](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/0f6851894ec09d11dee0b729da7a87430ec38ffb/scraper_news.py). That’s fine at the moment, but as we expand to have separate scrapers for each county, one file won’t be manageable.

We need:

- A directory to hold all the news-scraper-related code, e.g. `news/`
- A top-level python script to serve as an entry point to each of the scrapers. e.g. something like:

    ```sh
    # Run the SF scraper:
    > python3 scrape_news.py san_francisco

    # Run the Alameda scraper:
    > python3 scrape_news.py alameda

    # Run multiple scrapers:
    > python3 scrape_news.py san_francisco alameda
    ```

Mr0grog:
In order to make a big set of scrapers more manageable, this creates a `news` package where we can have one module per scraper (plus other support modules). The same script as before is the entry point, but you specify counties as arguments, e.g:

```sh
# Get San Francisco News
> python3 scraper_news.py san_francisco

# Get Alameda News
> python3 scraper_news.py alameda

# Get Both
> python3 scraper_news.py san_francisco alameda
```

If you don't specify any arguments it currently defaults to `san_francisco`, but in the future this will eventually be turned into a combined feed of all counties instead.

~Before merging, I want to clean up the argument parsing by using either `click` or `ArgParse`, but this should otherwise be good to go.~ *(This is done now.)*

Fixes #41.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 env/
 __pycache__/
+.mypy_cache/
 .DS_Store
 .vscode
--- a/news/__init__.py
+++ b/news/__init__.py
@@ -1,0 +1,14 @@
+from .san_francisco import SanFranciscoNews
+
+
+scrapers = {
+    'alameda': None,
+    'contra_costa': None,
+    'marin': None,
+    'napa': None,
+    'san_francisco': SanFranciscoNews,
+    'san_mateo': None,
+    'santa_clara': None,
+    'solano': None,
+    'sonoma': None,
+}
--- a/news/__init__.py
+++ b/news/__init__.py
@@ -1,8 +1,10 @@
+from typing import Dict, Type
 from .alameda import AlamedaNews
+from .base import NewsScraper
 from .san_francisco import SanFranciscoNews
 
 
-scrapers = {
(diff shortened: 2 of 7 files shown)


Mr0grog:
@ldtcooper this should be ready for review and merge. For the most part, it’s just rearranging things into separate folders and files, plus adding some docstrings (I’m using the [Numpy style](https://numpydoc.readthedocs.io/en/latest/format.html), but let me know if you want to use something different.)

The main functionality change here is using `click` to handle CLI arguments. I used it instead of `ArgParse` since it was already included in `requirements.txt`, but let me know if I should switch to something else.

Mr0grog:
> For the sake of consistency, can we use type hints in here?

Ah, definitely! I should have done that. Will add later today.

Mr0grog:
@ldtcooper added type hints. (I’ll add CI in a separate PR.) Anything else here?

PR #49, opened 2020-05-07
Support json_feed, json_simple, and rss news formats

### LINKED ISSUE #40: Add support for JSON feed for news
The news scraper currently outputs a custom, simplified news format like:

```js
{
  "newsItems": [
    {
      "url": "https://sf.gov/news/expansion-coronavirus-testing-all-essential-workers-sf",
      "text": "Expansion of coronavirus testing for all essential workers in SF",
      "date": "2020-04-23T04:11:56Z"
    },
    {
      "url": "https://sf.gov/news/sfmta-starts-closing-some-sf-streets-promote-safety-and-physical-distancing",
      "text": "SFMTA starts closing some SF streets to promote safety and physical distancing",
      "date": "2020-04-21T21:05:46Z"
    },
    // etc...
}
```

But we should really output a more standardized format. JSON feed (https://jsonfeed.org/) is a JSON-based alternative to RSS that a lot of news readers support.

Because the front-end will also need to transition to consume the new format, we need to have a CLI argument to choose the format, e.g:

```sh
# JSON Feed
> python3 scraper_news.py --format json

# Current JSON style
> python3 scraper_news.py --format json-simple
```

The JSON Feed format should be the default if `--format` is not specified.

### LINKED ISSUE #42: Add support for RSS/Atom for news
The news scraper currently outputs a custom, simplified JSON format, but we should really output either RSS or Atom — a standardized format so other tools and feed readers can use our output.

We need to have a CLI argument to choose the format, e.g:

```sh
# JSON Feed
> python3 scraper_news.py --format rss
```

See #40 about adding another format — JSON Feed — which will be the default.

Mr0grog:
You can now specify the output format for the news scraper with the `--format` option, which can be:
- `json_simple` (default) the current format
- `json_feed` the JSONFeed standard (https://jsonfeed.org/version/1)
- `rss` for RSS 2.0

Eventually, we'll switch `json_feed` to be the default, but for now we need to not break existing automation.

This uses dataclasses for some of its work, which are new in Python 3.7. I think that’s our minimum version, so this should be OK. (Is that right, @ldtcooper?)

You can also specify more than one format, and specify an output directory. That way, a command like this:

```sh
$ python scraper_news.py san_francisco alameda --format json_feed --format rss --output feed_data
```

Can write out 4 files, which would be a pain to deal with on STDOUT:
- `feed_data/san_francisco.json`
- `feed_data/san_francisco.rss`
- `feed_data/alameda.json`
- `feed_data/alameda.rss`

Fixes #40, fixes #42.

### DIFF
--- a/news/base.py
+++ b/news/base.py
@@ -1,5 +1,5 @@
 import requests
-from typing import List
+from typing import Dict, List
 from .feed import NewsFeed, NewsItem
 
 
@@ -19,8 +19,11 @@ class NewsScraper:
     scraping should start, then implement `parse_page()`, which returns a list
     of news items given some HTML.
     """
-    FEED_TITLE = ''
+    FEED_INFO: Dict = {}
     START_URL = ''
+
+    def create_feed(self) -> NewsFeed:
+        return NewsFeed(**self.FEED_INFO)
 
     def scrape(self) -> NewsFeed:
         """
@@ -38,7 +41,7 @@ class NewsScraper:
           'text': 'Expansion of coronavirus testing for all essential workers in SF',
           'date': '2020-04-23T04:11:56Z'}]
         """
-        feed = NewsFeed(title=self.FEED_TITLE)
+        feed = self.create_feed()
         # TODO: we may want to iterate through multiple pages in the future
         response = requests.get(self.START_URL)
         response.raise_for_status()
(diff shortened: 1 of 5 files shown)


Mr0grog:
D’oh! Just fixed the linting errors I should have caught locally.

Mr0grog:
@ldtcooper I went ahead and added RSS support and file output support here since nobody has reviewed. Do you have time give it a once-over?

Mr0grog:
Sweet, going ahead and merging this, then. Thanks!

Issue #46, opened 2020-05-02
Data: Create CLI to run county data scrapers. 


elaguerta:

elaguerta:
https://github.com/sfbrigade/data-covid19-sfbayarea/pull/34/files#r417520425

PR #51, opened 2020-05-13
#46 Add CLI to run data scrapers

elaguerta:
Add `scraper_data.py` and `run_scraper_data.sh` to mirror the news side. 
A few high-level questions:
1. `scraper_data.py` writes a file `data.json` in the main repo. Where should this file live? Should I add it to .gitignore?
2. Should we modify the `run_scraper_data.sh` script to accept command line arguments (specific counties)?
3. The news scraper has its own base class. Should I set up a parallel class for the data scrapers? The counties all seem like snowflakes. The only common method I would enforce for each scraper is`get_county()`, returning the top-level .json for that county

### DIFF
--- a/data_scrapers/alameda_county.py
+++ b/data_scrapers/alameda_county.py
@@ -30,7 +30,7 @@ def get_county() -> Dict:
     # Load data model template into a local dictionary called 'out'.
     with open('./data_models/data_model.json') as template:
         out = json.load(template)
-    
+
     # populate dataset headers
     out["name"] = "Alameda County"
     out["source_url"] = landing_page
@@ -41,7 +41,7 @@ def get_county() -> Dict:
     response.raise_for_status()
     cases_header = response.json()
     timestamp = cases_header["editingInfo"]["lastEditDate"]
-    # Raise an exception if a timezone is specified. If "dateFieldsTimeReference" is present, we need to edit this scrapr to handle it. 
+    # Raise an exception if a timezone is specified. If "dateFieldsTimeReference" is present, we need to edit this scrapr to handle it.
     # See: https://developers.arcgis.com/rest/services-reference/layer-feature-service-.htm#GUID-20D36DF4-F13A-4B01-AA05-D642FA455EB6
     if "dateFieldsTimeReference" in cases_header["editingInfo"] or "editFieldsInfo" in cases_header:
         raise FutureWarning("A timezone may now be specified in the metadata.")
@@ -61,7 +61,7 @@ def get_county() -> Dict:
 
 
 # Confirmed Cases and Deaths
-def get_timeseries() -> Dict: 
+def get_timeseries() -> Dict:
     """Fetch daily and cumulative cases and deaths by day
     Returns the dictionary value for "series": {"cases":[], "deaths":[]}.
     To create a DataFrame from this dictionary, run
@@ -79,12 +79,12 @@ def get_timeseries() -> Dict:
     }
 
     # query API
-    param_list = {'where':'0=0', 'resultType': 'none', 'outFields': 'Date,AC_Cases,AC_CumulCases,AC_Deaths,AC_CumulDeaths', 'outSR': 4326,'orderByField': 'Date', 'f': 'json'}
-    response = requests.get(cases_deaths, params=param_list)  # type: ignore
+    param_list = {'where':'0=0', 'resultType': 'none', 'outFields': 'Date,AC_Cases,AC_CumulCases,AC_Deaths,AC_CumulDeaths', 'outSR': '4326','orderByField': 'Date', 'f': 'json'}
+    response = requests.get(cases_deaths, params=param_list)
     response.raise_for_status()
     parsed = response.json()
     features = [obj["attributes"] for obj in parsed['features']]
-    
+
     # convert dates
     for obj in features:
         month, day, year = obj['Date'].split('/')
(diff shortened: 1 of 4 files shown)


ldtcooper:
To answer your questions as best I can: 
1. I think that we do want to write `data.json` to the root directory of the project. All that has to do is be easily accessible for the frontend team to input into the front end
2. I don't think there's a super strong need for that. If we want to add that, it should be lower priority
3. I agree that the counties are different enough that there's not much of a reason to set up a base class for the scrapers

Mr0grog:
>> 1. `scraper_data.py` writes a file `data.json` in the main repo. Where should this file live? Should I add it to .gitignore?
> 
> I think that we do want to write `data.json` to the root directory of the project. All that has to do is be easily accessible for the frontend team to input into the front end

Can I suggest *not* writing to disk by default, and accepting an `--output <path>` parameter that will cause it write at a chosen location? That’s how I’ve structured news and I think it’s a good pattern: https://github.com/sfbrigade/data-covid19-sfbayarea/blob/f8b7923d07ea9f4d961215f77f73245334796983/scraper_news.py#L19-L20

and: https://github.com/sfbrigade/data-covid19-sfbayarea/blob/f8b7923d07ea9f4d961215f77f73245334796983/scraper_news.py#L40-L46

(In this case, the output path is for a directory, because it may write *several* files, depending on the other args you’ve provided.)

Mr0grog:
Agree on not needing a base class — it’s really only valuable if you think there’s important shared behavior or if you need to store state. Otherwise, functions in a module are 100% A-OK. And I don’t think the data scrapers have the kind of clear patterns or state that would benefit from a class yet. 👍

data_scrapers/alameda_county.py
```diff
@@ -144,12 +144,13 @@ def get_demographics(out:Dict) -> (Dict, List):
     # format query to get entry for Alameda County
     param_list = {'where': "Geography='Alameda County'", 'outFields': '*', 'outSR':4326, 'f':'json'}
     # get cases data
-    response = requests.get(demographics_cases, params=param_list)
+    response = requests.get(demographics_cases, params = param_list) # type: ignore
```
Mr0grog, line 147: It would be better if we could resolve this without bailing out on types. Some options that should work:

- Make `outSR` a string:

    ```py
    param_list = {'where': "Geography='Alameda County'", 'outFields': '*', 'outSR': '4326', 'f':'json'}
    ```

- Define `param_list` as a compatible type. Unfortunately the actual type definition is [absurdly complex](https://github.com/python/typeshed/blob/ad6cf3a314fbecc5df7244664bb245870bae5552/third_party/2and3/requests/api.pyi#L14-L24), so we shouldn’t duplicate it, but we could do:

    ```py
    param_list: Dict[str, Union[str, int]] = {'where': "Geography='Alameda County'", 'outFields': '*', 'outSR': 4326, 'f':'json'}
    ```

    (Since we have more than one kind of value, MyPy just throws up its hands and says “could be anything,” which is too broad for the actual definition, so it needs a little hint here.)

data_scrapers/alameda_county.py
```diff
@@ -80,7 +80,7 @@ def get_timeseries() -> Dict:
 
     # query API
     param_list = {'where':'0=0', 'resultType': 'none', 'outFields': 'Date,AC_Cases,AC_CumulCases,AC_Deaths,AC_CumulDeaths', 'outSR': 4326,'orderByField': 'Date', 'f': 'json'}
-    response = requests.get(cases_deaths, params=param_list)
+    response = requests.get(cases_deaths, params=param_list)  # type: ignore
```
Mr0grog, line 83: Same comment as the other `requests.get()` call with respect to types.

data_scrapers/alameda_county.py
```diff
@@ -144,12 +144,13 @@ def get_demographics(out:Dict) -> (Dict, List):
...
     # get cases data
-    response = requests.get(demographics_cases, params=param_list)
+    response = requests.get(demographics_cases, params = param_list) # type: ignore
     response.raise_for_status()
     parsed = response.json()
     cases_data = parsed['features'][0]['attributes']
     # get deaths data
-    response = requests.get(demographics_deaths, params=param_list)
+    response = requests.get(demographics_deaths,
+                            params=param_list)  # type: ignore
```
Mr0grog, line 153: Same comment as the other `requests.get()` call with respect to types.

Mr0grog:
(Side note: if your editor has a [plugin for EditorConfig](https://editorconfig.org/), you can install it and it’ll automatically clean up some of the spaces at the ends of lines and so on.)

elaguerta:
@Mr0grog @ldtcooper Thanks for your comments. Ready for another round of review. 
(PS sorry about the multiple failed checks on circleci - it took me a minute before I thought to figure out how run MyPy locally!)

elaguerta:
@Mr0grog This commit touches the alameda scraper. I do plan to address your comment [comment on#34 about making a util for getting a data template object](https://github.com/sfbrigade/data-covid19-sfbayarea/pull/34#discussion_r420591276). I'll roll it into PR #52.

run_scraper_data.sh
```diff
@@ -0,0 +1,2 @@
+source env/bin/activate;
+python3 scraper_data.py; # how do we pass command line args to python?
```
Mr0grog, line 2: FYI, you can pass all arguments along via:

```suggestion
python3 scraper_data.py $@;
```

(In bash, `$@` is a list of all arguments, `$#` is the number of arguments, and `$1`, `$2`, etc. are the individual arguments.)

PR #52, opened 2020-05-13
Refactor SF county

elaguerta:
Refactor SF county data fetcher to align with data model


data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
+    # Dict of target_label : source_label for re-keying.
+    # Note: Native_Amer not currently reported
+    RACE_ETH_KEYS = {"Latinx_or_Hispanic": "Hispanic or Latino", "Asian": "Asian", "African_Amer": "Black or African American",
+                 "White": "White", "Pacific_Islander": "Native Hawaiian or Other Pacific Islander", "Multiple_Race": "Multiple Race",
+                 "Other": "Other", "Unknown": "Unknown"}
+    data = response.json()
+    # re-key and aggregate to flatten race x ethnicity
+    race_eth_data: Dict[str,int] = { k:0 for k in RACE_ETH_KEYS.keys() } # initalize all categories to 0 for aggregating
+
+    for item in data: # iterate through all race x ethnicity objects
```
elaguerta, line 209: I'd like some more eyeballs on this logic. I'm getting numbers close to how SF county flattens race and ethnicity on their dashboard, but not exact. The discrepancy might be the lag between the data update and the dashboard update, but then you would expect our numbers to be slightly higher. However, they are slightly lower for Hispanic or Latino, and Other. I think Jason Lally mentioned that they are revisiting their data categories. Perhaps some "Unknown" and "Other" patients are being re-classified, explaining the drop in numbers in those categories today compared to the dashboard numbers. 

Here's what this for loop produces:
```
   "race_eth": {
            "Latinx_or_Hispanic": 828,
            "Asian": 257,
            "African_Amer": 98,
            "White": 343,
            "Pacific_Islander": 23,
            "Multiple_Race": 10,
            "Other": 99,
            "Unknown": 336
        }
```

Here's what the dashboard looks like (updated 5/12):
Screen Shot 2020-05-13 at 8 53 20 AM

elaguerta: Argh I just noticed that Native American isn't listed in the raw data. It looks like "race" is blank for the entry with 6 cases - that could possibly be Native American. Time to slack Jason Lally!

kwonangela7: Re: unknown, I think you might be missing the case where Race = Unknown and Ethnicity = Not Hispanic or Latino. That would surface the 12 you're missing to get the right number of 348.

Re: Hispanic or Latino - I'm not sure. Is it possible that the dashboard is wrong? Or my understanding of what should fall into this category is wrong? It makes sense to me that only the first three rows of the raw data set should be added (and your logic does that), and that only adds up to 828. I'll revisit this later tonight or tomorrow.

Re: White - I didn't have as much time to look at this one, so I'll try to look at it again tomorrow, but I think you're getting a higher count b/c you're counting (Race: White) + (Ethnicity: Hispanic or Latino) as White, but I think it's already accounted for in Hispanic or Latino?

kwonangela7: I don't think the Hispanic or Latino discrepancy can be explained by this (the discrepancy seems too large), but just posting this disclaimer from the raw data set anyway: "Unknowns may become known over time as more information is gathered. Combinations of race and ethnicity resulting in 5 or fewer cases are dropped for privacy reasons.", in case it happens to be useful at some point.

data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
+
+        # handle unknown
+        # sum over only items for which BOTH race and ethnicity are Unknown, or one is Unknown and the other is not reported
+        if "race" in cols and "ethnicity" in cols:
+            if item["race"] == "Unknown" and item["ethnicity"] == "Unknown": # both race and ethnicity are Unknown
+                race_eth_data["Unknown"] += cases
+        elif "ethnicity" in cols and item["ethnicity"] == "Unknown": # or, race not reported and ethnicity is unknown
+            race_eth_data["Unknown"] += cases
+        elif "race" in cols and item["race"] == "Unknown": # or, ethnicity not reported and race is unknown
+            race_eth_data["Unknown"] += cases
```
Mr0grog, line 221: Is there a meaningful difference between “not reported” and “unknown”? It doesn’t seem like it in the logic here, so you could probably simplify and make it easier to follow with:

```py
race = item.get('race', 'Unknown')
ethnicity = item.get('ethnicity', 'Unknown')

# Now you always have a value for both, so there's no need for the complex
# branching, and no need for the `if 'race' in cols and ...` checks:
if race == 'Unknown' and ethnicity == 'Unknown':
    race_eth_data['Unknown'] += cases

if ethnicity == 'Hispanic or Latino':
    race_eth_data["Latinx_or_Hispanic"] += cases

# etc.
```

data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
+        
+        # sum over all known races
+        if "race" in cols and item["race"] != "Unknown": 
+            if item["race"] == "Other":  # except for race = Other, ethnicity = Hispanic or Latino; exclude Other/Hispanic Latino from Other
+                if "ethnicity" in cols and item["ethnicity"] != "Hispanic or Latino": 
+                    race_eth_data["Other"] += cases
+            else:
+                for k,v in RACE_ETH_KEYS.items(): # look up this item's re-key
+                    if v == item["race"]:
+                        race_eth_data[k] += cases
```
Mr0grog, line 235: Are we double-counting when `ethnicity == 'Hispanic or Latino'` here? I think `{race: 'Unknown', ethnicity: 'Hispanic or Latino'}` and `{race: 'White', ethnicity: 'Hispanic or Latino'}` are both in the data and would wind up getting here AND to line 225, but your screenshot from the dashboard shows totals that sum up to 100%.

data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
-    positives = get_json(tests_url + total_positives_query)
-    tests[0].update(positives[0])
-    return tests
+    response = requests.get(race_ethnicity_url)
+    response.raise_for_status()
+    # Dict of target_label : source_label for re-keying.
+    # Note: Native_Amer not currently reported
+    RACE_ETH_KEYS = {"Latinx_or_Hispanic": "Hispanic or Latino", "Asian": "Asian", "African_Amer": "Black or African American",
+                 "White": "White", "Pacific_Islander": "Native Hawaiian or Other Pacific Islander", "Native_Amer": "Native American", "Multiple_Race": "Multiple Race",
+                 "Other": "Other", "Unknown": "Unknown"}
```
Mr0grog, line 204: If you inverted the keys and values here, you wouldn’t need the loop on line 233 and none of your other code would get more complex (you just need to trade `RACE_ETH_KEYS.keys()` for `RACE_ETH_KEYS.values()` on line 207 below).

data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
+
+RESOURCE_IDS = {'cases_deaths_transmission': 'tvq9-ec9w', 'age_gender': 'sunc-2t3k', 
+                'race_eth': 'vqqm-nsqg', 'tests': 'nfpa-mg4g' }
+
+metadata_url = 'https://data.sfgov.org/api/views/metadata/v1/'
+data_url = 'https://data.sfgov.org/resource/'
+age_gender_url = f"{data_url}{RESOURCE_IDS['age_gender']}.json"
+race_ethnicity_url = f"{data_url}{RESOURCE_IDS['race_eth']}.json"
+transmission_url = f"{data_url}{RESOURCE_IDS['cases_deaths_transmission']}.json"
+tests_url = f"{data_url}{RESOURCE_IDS['tests']}.json"
```
Mr0grog, line 17: This isn’t a huge deal, but you could probably reduce some boilerplate *and* make your code a bit more readable by wrapping this up into a class for dealing with the API, e.g:

```py
class SanFranciscoApi:
    base_url = 'https://data.sfgov.org/'

    def __init__(self):
        # Having a session gets you connection pooling and other
        # performance niceties.
        self.session = requests.Session()
        self.resource_url = urljoin(self.base_url, '/resource/')
        self.metadata_url = urljoin(self.base_url, '/api/views/metadata/v1/')

    def request(self, url, **kwargs):
        response = self.session.request(url, **kwargs)
        response.raise_for_status()
        return response.json()

    def get_cases(self, params):
        return self.request(f'{self.resource_url}tvq9-ec9w', params)

    def get_age_and_gender(self, params):
        return self.request(f'{self.resource_url}sunc-2t3k', params)

    # etc.

# Now in your functions you can have a slightly more readable call
# and not repeat all the business around checking for exceptions and
# parson JSON:
cases_series = api_instance.get_cases({'case_disposition': 'Confirmed',
                                       '$select':'date,sum(case_count) as cases',
                                       '$group':'date',
                                       '$order':'date'})
```

And this gives you more places to do useful other things, like setting a common user-agent string for all your requests.

If any other counties use Socrata, you can split this in two so parts of it are also reusable:

```py
class SocrataApi:
    def __init__(self):
        # Having a session gets you connection pooling and other
        # performance niceties.
        self.session = requests.Session()
        self.resource_url = urljoin(self.base_url, '/resource/')
        self.metadata_url = urljoin(self.base_url, '/api/views/metadata/v1/')

    def request(self, url, **kwargs):
        response = self.session.request(url, **kwargs)
        response.raise_for_status()
        return response.json()

class SanFranciscoApi(SocrataApi):
    base_url = 'https://data.sfgov.org/'

    def get_cases(self, params):
        return self.request(f'{self.resource_url}tvq9-ec9w', params)

    def get_age_and_gender(self, params):
        return self.request(f'{self.resource_url}sunc-2t3k', params)

    # etc.
```

data_scrapers/san_francisco_county.py
elaguerta: Yes we were double counting for White, Hispanic or Latino as @kwonangela7 [notes](https://github.com/sfbrigade/data-covid19-sfbayarea/pull/52#discussion_r424857992). For any other race and Hispanic or Latino, I believe the individual gets counted for both their race and as Hispanic Latino. I guess SF excludes White & Hispanic or Latino from White. I'll edit this so it lines up with SF.

data_scrapers/san_francisco_county.py
```diff
@@ -4,101 +4,238 @@
...
+
+def get_county() -> Dict:
+    """Main method for populating county data .json"""
+
+    # Load data model template into a local dictionary called 'out'.
+    with open('./data_models/data_model.json') as template:
+        out = json.load(template)
```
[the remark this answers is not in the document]
elaguerta: See data_scrapers/utils.py

data_scrapers/san_francisco_county.py
Mr0grog: Ah, sorry, I totally missed that last paragraph in her comment!

I guess I was reacting to the column in your screenshot being labeled “percent of cases” and noticing that they add up to 100%. That implies we shouldn’t double-count like this, although I agree that seems unexpected — if you’re doing anything other than just showing each (race × ethnicity) combination, you *should* get > 100%. 😕

> I'll edit this so it lines up with SF.

I didn’t follow the conversations about this when creating the data model closely, so sorry if this is already well-trodden ground: given the confusion you’re encountering around reverse-engineering SF’s logic here, is matching even what we actually want? If so, 👍

data_scrapers/san_francisco_county.py
elaguerta: Some progress on this. Here's our data as of 5/17:
```
 "race_eth": {
        "Latinx_or_Hispanic": 881,
        "Asian": 263,
        "African_Amer": 105,
        "White": 323,
        "Pacific_Islander": 25,
        "Native_Amer": 8,
        "Multiple_Race": 10,
        "Other": 107,
        "Unknown": 346
      }
```
Here's what the dashboard is showing, updated 5/16:
Screen Shot 2020-05-17 at 9 55 04 PM

Here are the major changes I made to the logic:
  1. Include Race = 'Unknown', Ethnicity = 'Not Hispanic or Latino' in our 'Unknown'
2. Exclude Race = 'Other', Ethnicity = 'Hispanic or Latino' from our 'Other'
3. Exclude Race = 'White', Ethnicity = 'Hispanic or Latino' from our 'White
4. If Race was not reported, counted it the value as Race = 'Native American' (!!!)
This last one is because the dashboard has Native American, but the [raw data](https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Race-and-Ethnicity/vqqm-nsqg/data) doesn't, however, the one value in the raw data that doesn't have race reported looks to be very close to 'Native American'. 
I am adding this very heavy-handed interpretation to our metadata.
Thoughts? @kwonangela7 @Mr0grog

elaguerta:
@Mr0grog ready for another round of review. I added a utils file, which has a function for getting the data model and a SocrataApi class. I also reformatted the age tables to match the README.

data_scrapers/san_francisco_county.py
Mr0grog: > This last one is because the dashboard has Native American, but the raw data doesn't, however, the one value in the raw data that doesn't have race reported looks to be very close to 'Native American'.

Ahhhhhhh! Yeah, that makes sense, although, ugh, I definitely hear you on how making this assumption seems like a bad thing. Definitely worth asking the county, but I also wouldn’t block merging while waiting for that feedback (we can always make changes in the future when we hear back).

I also think 2 & 3 feel weird, or at least not what I’d expect based on how categories are named. and I think the best way to characterize this is that, even though our categories are cross-cutting (i.e., based on their name, someone could fit in more than one), we are ensuring every case only gets bucketed in *one* (so they sum up to 100% of cases, rather than > 100%).

**I don’t have a strong feel on whether doing this in order to match SF’s dashboard is good. If we do it, we should add a note about it to `meta_from_baypd`.** It might be good to add a big comment at the top of `get_race_eth_table()` that describes how cases are getting bucketed, similar to what you wrote here (that way someone can get a sense of it without needed to puzzle through the actual logic).

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,268 @@
...
+        return out_series
+
+    # Confirmed Cases and Deaths by Date and Transmission
+    # Note that cumulative totals are not directly reported, we are summing over the daily reported numbers
+    def get_cases_series(self) -> List[Dict]:
+        """Get cases timeseries json, sum over transmision cat by date"""
+        resource_id = self.resource_ids['cases_deaths_transmission']
+        params = {'case_disposition': 'Confirmed',
+                '$select': 'date,sum(case_count) as cases', '$group': 'date', '$order': 'date'}
+        data = self.request(resource_id, params=params)
```
Mr0grog, line 106: Ohhh, this is not really what I had in mind when I recommended a `SanFranciscoApi` class **(although it’s also totally fine; no need to do a whole bunch more refactoring at this point if it’s getting tedious)**.

I was trying to separate the concerns around “get me some raw data from the API” and “parse/reformat/reorganize the API’s data.” So you might have:

```py
# These functions are purely about requesting data from a remote API and
# deal only with network and parse-related things (e.g. determining URLs,
# retrying, parsing JSON, caching responses, etc.).
class SanFranciscoApi(SocrataApi):
    ...

    def get_cases_deaths_transmission(self, **kwargs):
        """
        Get the total cases by date, transmission type and disposition
        ('Death' or 'Confirmed').
        Browse this at: https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Date-Transmission-and/tvq9-ec9w
        """
        return self.request('tvq9-ec9w', **kwargs)

    # Or, a slightly more complex version that might simplify your usage.
    # Not sure if there are other requests to tvq9-ec9w that would be more
    # complex because of this.
    def get_cases_deaths_transmission(self, disposition, **kwargs):
        """
        Get the total cases by date, and transmission type for a given
        disposition ('Death' or 'Confirmed').
        Browse this at: https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Date-Transmission-and/tvq9-ec9w
        """
        params = {'case_disposition': disposition,
                  '$select': 'date,sum(case_count) as cases',
                  '$group': 'date',
                  '$order': 'date'}
        return self.request('tvq9-ec9w', params=params, **kwargs)


# In the same set of basic functions you had before:
def get_cases_series(api):
    data = api.get_cases_deaths_transmission({
        'case_disposition': 'Confirmed',
        '$select': 'date,sum(case_count) as cases',
        '$group': 'date',
        '$order': 'date'
    })
    # Or the fancier version:
    data = api.get_cases_deaths_transmission('Confirmed')

    # convert date from ISO string to 'yyyy-mm-dd'. convert number strings
    # to int. calculate daily cumulative.
    cases_series = []
    cumul = 0
    for entry in data:
        entry["date"] = entry["date"][0:10]
        entry["cases"] = int(entry["cases"])
        cumul += entry["cases"]
        entry["cumul_cases"] = cumul
        cases_series.append(entry)
    return cases_series
```

That probably doesn’t look like it’s saving you a whole lot, but it’s still pretty useful. You get function names that more clearly tell someone else reading your code just what’s happening, a better place to document issues around SF’s API itself, and a clear place to do other things that are particular to the API, like special parameter parsing, caching, HTTP headers, etc.

If you have a situation like Alameda, where you have both an API and some dashboards you’re scraping, you might have two separate clients to encapsulate the different concerns around making API requests vs. managing a webdriver instance.

The way you have it now, I’m not sure there’s a huge benefit from being methods on a `SanFranciscoApi` class rather than just creating an instance of `SocrataApi` and passing it around:

```py
def get_county():
    ...
    session = SocrataApi('https://data.sfgov.org/')
    ...
    get_cases_series(session)
    ...
```

(Having the class means you don’t have to pass the instance around, which is useful, but not a huge benefit.)

Anyway, what you have here is OK, too! Separating it out a bit more can help others understand or modify your code a bit more easily (or help you do so a couple weeks from know when you’ve forgotten all the details of what you did).

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,268 @@
...
+        Note that SF does not provide death totals, so these datapoints will be -1.
+        To crete a DataFrame from the dictionary, run 'pd.DataFrame(get_demographics())'
+        """
+        # copy dictionary structure of global 'out' dictionary to local variable
+        demo_totals: Dict[str,Dict] = { "case_totals": dict(), "death_totals": dict() }
+        demo_totals["case_totals"]["gender"] = self.get_gender_table()
+        demo_totals["case_totals"]["age_group"] = self.get_age_table()
+        demo_totals["case_totals"]["transmission_cat"] = self.get_transmission_table()
+        demo_totals["case_totals"]["race_eth"] = self.get_race_eth_table()
+        return demo_totals
```
Mr0grog, line 84: I’ve lost track in this refactor: did we previously fill in `death_totals` for SF? It’s missing now.

(If no, it might good to add a `TODO` comment in the source calling out that this field needs work to fill in.)

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,268 @@
...
+        age_table : List[Dict[str,Any]] = [
+            {"group": "18_and_under", "raw_count": -1},
+            {"group": "18_to_30", "raw_count": -1},
+            {"group": "31_to_40", "raw_count": -1},
+            {"group": "41_to_50", "raw_count": -1},
+            {"group": "51_to_60", "raw_count": -1},
+            {"group": "61_to_70", "raw_count": -1},
+            {"group": "71_to_80", "raw_count": -1},
+            {"group": "81_and_older", "raw_count": -1}
+	    ]
```
Mr0grog, line 182: Kind of like I noted for Alameda county, you don’t need to write all this out if you make sure `AGE_KEYS` is in order.

data_scrapers/utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+class SocrataApi:
+    """ Class for starting a Session with Socrata endpoints and sending requests"""
+    def __init__(self, base_url):
+        self.session = requests.Session()
+        self.base_url = base_url
+        self.resource_url = urljoin(self.base_url, '/resource/')
+        self.metadata_url = urljoin(self.base_url, '/api/views/metadata/v1/')
+
+    def request(self, url, **kwargs):
```
Mr0grog, line 22: It might be good to add some caching here for cases where the same URL is requested multiple times (e.g. at least two metadata functions grab the same set of URLs).

The most quick-n-dirty built-in you could use is `functools.lru_cache`, which will leak memory, but probably isn’t a huge deal for our current use cases:

```py
import functools

class SocrataApi:
    @functools.lru_cache()
    def request(self, url, **kwargs):
        ...
```

The [methodtools package](https://pypi.org/project/methodtools/) has a version of `lru_cache` that won’t leak.

A better long-term approach might be to use [cachecontrol](https://cachecontrol.readthedocs.io), though. It’s designed more precisely for caching with `requests`.

data_scrapers/alameda.py
```diff
@@ -168,26 +178,44 @@ def get_demographics(out:Dict) -> (Dict, List):
         demo_totals["case_totals"]["race_eth"][k] = cases_data[v]
         demo_totals["death_totals"]["race_eth"][k] = deaths_data['Deaths_' + v]
     # get age cases and deaths
-    demo_totals["case_totals"]["age_group"] = { k: v for k, v in cases_data.items() if 'Age' in k }
-    demo_totals["death_totals"]["age_group"] = {k: v for k, v in deaths_data.items() if 'Age' in k}
+    # get age groups. Our data model calls for a list, but these may be out of age order.
+    demo_totals["case_totals"]["age_group"] = { k:v for k, v in cases_data.items() if 'Age' in k }
+    demo_totals["death_totals"]["age_group"] = { k:v for k, v in deaths_data.items() if 'Age' in k }
```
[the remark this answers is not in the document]
elaguerta: Somehow I made it this far without knowing that Python3 preserves dictionary ordering! This is the best news I've heard during quarantine. Low bar, but life-changing.

data_scrapers/san_francisco.py
elaguerta: We didn't have death_totals (and still don't). Jason Lally says it should be up this week. I'll add a `TODO` to fill in.

data_scrapers/alameda.py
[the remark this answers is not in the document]
Mr0grog: Important caveat: ordering isn’t involved when *comparing* dicts (for that, use `collections.OrderedDict`). But for iteration, it’s a *very* nice improvement over Python 2. 😄

data_scrapers/san_francisco_county.py
elaguerta: So I've decided to assign each individual to one bucket. I've add a description in `meta_from_baypd` and in comments at the top of get_race_eth_table().
I think is what our logic was doing anyways, I've just made it more clear. In any case I'm getting the same numbers as before with this approach - close to what's on the SF dashboard. 
For comparison, here's a [snapshot of Sonoma county's race/eth table](
https://socoemergency.org/emergency/novel-coronavirus/coronavirus-cases/), 
Screen Shot 2020-05-19 at 7 20 50 AM
It looks like they are bucketing each individual in one category, so the partition sums up to 100% of cases. Looks like they are only putting an individual in a race bucket if they are not Hispanic/Latino. Since Sonoma is doing it this way, and it looks like SF could be doing it this way, I'm inclined to do the same.

elaguerta:
@Mr0grog ready for another round of review. Some highlights: 
- san_francisco.py: deleted SanFranciscoApi class. Instead I'm passing around the SocrataApi instance. I decided not to separate the requests from the transforming/parsing in each method. Each method seems to be a reasonable enough length, and the queries and transform processes are pretty specific to each table. 
- utils.py: imported cachecontrol and created a cached session on the SocrataApi class. not sure if we need to keep the uncached session around

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+    """ Main method for populating county data.json """
+    # TODO: Add deaths data when it is available through the api.
+    # TODO: In get_race_eth_table(), update code when SF provides a race label for the Native American datapoint.
+
+    # Load data model template into a local dictionary called 'out'.
+    out = get_data_model()
+    # create a SocrataApi instance
+    RESOURCE_IDS = {'cases_deaths_transmission': 'tvq9-ec9w', 'age_gender': 'sunc-2t3k',
+                     'race_eth': 'vqqm-nsqg', 'tests': 'nfpa-mg4g'}
+    session = SocrataApi('https://data.sfgov.org/', RESOURCE_IDS)
```
Mr0grog, line 17: This setup works fine, but feels a bit weird to me. If I’m reading right, you’re really just putting a `resource_ids` attribute on `SocrataApi` so you can access it in other places here, and it actually has nothing to do with `SocrataApi` itself.

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+    for entry in data:
+        entry["date"] = entry["date"][0:10]
+        entry["cases"] = int(entry["cases"])
+        cumul += entry["cases"]
+        entry["cumul_cases"] = cumul
+        cases_series.append(entry)
+    return cases_series
+
+
+def get_deaths_series(session) -> List[Dict]:
```
Mr0grog, line 107:
```suggestion
def get_deaths_series(session: SocrataApi) -> List[Dict]:
```

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+        entry["date"] = entry["date"][0:10]
+        entry["deaths"] = int(entry["deaths"])
+        cumul += entry["deaths"]
+        entry["cumul_deaths"] = cumul
+        death_series.append(entry)
+    return death_series
+
+# Daily count of tests with count of positive tests
+# Note that SF county does not include pending tests, and does not directly report negative tests or cumulative tests.
+def get_tests_series(session) -> List[Dict]:
```
Mr0grog, line 127:
```suggestion
def get_tests_series(session: SocrataApi) -> List[Dict]:
```

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+                        positive=int(entry["pos"]))
+        out_entry['negative'] = out_entry["tests"] - out_entry["positive"]
+        totals.update(out_entry)
+        out_entry.update(cumul_tests=totals['tests'],
+                        cumul_pos=totals['positive'],
+                        cumul_neg=totals['negative'])
+        test_series.append(out_entry)
+    return test_series
+
+def get_age_table(session) -> List[Dict]:
```
Mr0grog, line 149:
```suggestion
def get_age_table(session: SocrataApi) -> List[Dict]:
```

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+    # flatten data into a dictionary of age_group:cases
+    data = { item["age_group"] : int(item["sum_confirmed_cases"]) for item in data }
+    age_table = []
+    # fill in values in age table
+    for target_key, source_key in AGE_KEYS.items():
+        age_table.append( { "group": target_key, "raw_count": data[source_key] })
+
+    return age_table
+
+def get_gender_table(session) -> Dict:
```
Mr0grog, line 169:
```suggestion
def get_gender_table(session: SocrataApi) -> Dict:
```

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+    # Note: non cis genders not currently reported
+    resource_id = session.resource_ids['age_gender']
+    GENDER_KEYS = {"Female": "female", "Male": "male", "Unknown": "unknown"}
+    params = {'$select': 'gender, sum(confirmed_cases)', '$group': 'gender'}
+    data = session.resource(resource_id, params=params)
+    # re-key
+    return {GENDER_KEYS[entry["gender"]]: int(entry["sum_confirmed_cases"])
+            for entry in data}
+
+def get_transmission_table(session) -> Dict:
```
Mr0grog, line 181:
```suggestion
def get_transmission_table(session: SocrataApi) -> Dict:
```

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,232 @@
...
+    TRANSMISSION_KEYS = {"Community": "community",
+                        "From Contact": "from_contact", "Unknown": "unknown"}
+    params = { '$select': 'transmission_category, sum(case_count)', '$group': 'transmission_category'}
+    data = session.resource(resource_id, params=params)
+    # re-key
+    transmission_data = { TRANSMISSION_KEYS[ entry["transmission_category"] ]: int(entry["sum_case_count"]) for entry in data}
+    return transmission_data
+
+# Confirmed cases by race and ethnicity
+def get_race_eth_table(session) -> Dict:
```
Mr0grog, line 194:
```suggestion
def get_race_eth_table(session: SocrataApi) -> Dict:
```

data_scrapers/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    with template_path.open() as template:
+        out = json.load(template)
+    return out
+
+class SocrataApi:
+    """
+    Class for starting a session for requests via Socrata APIs.
+    Initialize with a base_url and an optional dictionary of resource_ids = { key:id_string }
+    """
+    def __init__(self, base_url, resource_ids = None):
```
Mr0grog, line 20:
```suggestion
    def __init__(self, base_url: str, resource_ids: Dict[str, str] = None):
```

data_scrapers/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        self.session = requests.Session()
+        self.cached_sess = CacheControl(self.session)
+        self.base_url = base_url
+        self.resource_url = urljoin(self.base_url, '/resource/')
+        self.metadata_url = urljoin(self.base_url, '/api/views/metadata/v1/')
+        self.resource_ids = resource_ids
+
+    def request(self, url, **kwargs):
```
Mr0grog, line 28:
```suggestion
    def request(self, url: str, **kwargs: Any) -> Dict:
```

data_scrapers/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        response = self.cached_sess.get(url, **kwargs)
+        response.raise_for_status()
+        return response.json()
+
+    def resource(self, resource_id, **kwargs) -> Dict:
```
Mr0grog, line 33:
```suggestion
    def resource(self, resource_id: str, **kwargs: Any) -> Dict:
```

data_scrapers/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        return self.request(f'{self.resource_url}{resource_id}', **kwargs)
+
+    def metadata(self, resource_id, **kwargs) -> Dict:
```
Mr0grog, line 36:
```suggestion
    def metadata(self, resource_id: str, **kwargs: Any) -> Dict:
```

Mr0grog:
> utils.py: imported cachecontrol and created a cached session on the SocrataApi class. not sure if we need to keep the uncached session around

I don’t think there’s any reason to keep a reference to the uncached version of the session.

data_scrapers/alameda.py
```diff
@@ -154,40 +153,58 @@ def get_demographics(out: Dict) -> Tuple[Dict, List]:
...
+    demo_totals = {
+        "case_totals": out["case_totals"], "death_totals": out["death_totals"]}
 
-    # Parse and re-key
+    # Parse and re-key demo_totals
     # gender cases and deaths
     for k, v in GENDER_KEYS.items():
         demo_totals["case_totals"]["gender"][k] = cases_data[v]
-        if k in deaths_data.keys(): # the deaths table does not currently include MTF or FTM
+        if 'Deaths_' + v in deaths_data.keys():  # the deaths table does not currently include MTF or FTM
```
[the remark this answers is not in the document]
elaguerta: I regret all this time I've been missing out on hashed lookup. Thanks!

data_scrapers/san_francisco.py
elaguerta: Yes. I can see now why it's weird. Should I take the `resource_ids` off of the `SocrataApi` instance? I can just as easily pass them around along with the `session`.

elaguerta: Ok, I ended up taking it off  the `SocrataApi` because it is way easier to do type hints. I begrudgingly admit that this type hinting thing might be forcing me to think more rigorously.

elaguerta:
@Mr0grog Ready for review. I checked in with Jason, sounds like Deaths data is coming. 

@kwonangela7 after all my hand-wringing over the race x eth flattening, [they have collapsed the categories for us](https://data.sfgov.org/COVID-19/COVID-19-Cases-Summarized-by-Race-and-Ethnicity/vqqm-nsqg/data).

data_scrapers/alameda.py
```diff
@@ -154,40 +153,60 @@ def get_demographics(out: Dict) -> Tuple[Dict, List]:
     parsed = response.json()
     deaths_data = parsed['features'][0]['attributes']
 
+    # join cases and deaths tables in a temporary dictionary, to use for checking for values <10
+    demo_data = {"case_totals": cases_data, "death_totals": deaths_data}
+
+    # Handle values equal to '<10', if any. Note that some data points are entered as `null`, which
+    # will be decoded as Python's `None`
+    # TODO: As of 5/23/20, there are no string values "<10" in the source data. The dashboard disclaimers list categories with counts < 10 that have been supressed. Consider eliminating the code dealing with "<10".
```
Mr0grog, line 161: 👍

data_scrapers/san_francisco.py
Mr0grog: Sorry I wasn’t able to reply yesterday, but this seems good! 😅

data_scrapers/san_francisco.py
```diff
@@ -0,0 +1,222 @@
...
+    # calculate daily cumulative
+    cases_series = []
+    cumul = 0
+    for entry in data:
+        entry["date"] = entry["date"][0:10]
+        entry["cases"] = int(entry["cases"])
+        cumul += entry["cases"]
+        entry["cumul_cases"] = cumul
+        cases_series.append(entry)
+    return cases_series
```
Mr0grog, line 104: I don’t think this is a blocker, but I just noticed we are creating time series with gaps here. e.g. I’m getting:

```
{
    "date": "2020-03-05",
    "cases": 2,
    "cumul_cases": 2
},
{
    "date": "2020-03-06",
    "cases": 5,
    "cumul_cases": 7
},
{
    "date": "2020-03-08",
    "cases": 6,
    "cumul_cases": 13
},
...
```

It looks like there are simply gaps in the source data (probably on days where there were 0 new cases/deaths), so it’s not something we’re doing *wrong* here.

If that’s not a serious issue, let’s not worry about it for this PR (and if we want to “fix” it, fix it as a separate PR).

This issue also applies to the other time series.

elaguerta: Sounds good, I'll file this as a separate issue. Thanks for catching these.

PR #54, opened 2020-05-14
Add executable bit to scripts

rickpr:
Change mode on the scripts to make it more clear that these will be run from the terminal after cloning.


Mr0grog:
Not sure why this been sitting for so long. I’m rebasing it, adding the executable bit to scripts that didn’t exist when this was first created, and merging.

Issue #48, opened 2020-05-05
Data: Update data_scrapers/README.md 


elaguerta:
A list in no particular order of things to explain and comment on: 
Data cleaning:
1. Add data points with abnormalities like '<10' to metadata
2. convert strings to int
3. look for timezone info
4. If cumulative numbers are reported per day, look for a disclaimer

Methodology for collapsing race/ethnicity:
From SF:
"In the race/ethnicity data shown below, the "Other” category includes those who identified as Other or with a race/ethnicity that does not fit the choices collected. The “Unknown” includes individuals who did not report a race/ethnicity to their provider, could not be contacted, or declined to answer."
Here's a possible workflow:
1. Sum over race categories, except for unknown Race.  
2. Sum over Hispanic/Latino (all races) - note that this will double-count individuals who have a known race and are Hispanic/Latino
3. Set our "unknown" datapoint to count only individuals with **both** unknown race & unknown ethnicity.

Issue #29, opened 2020-04-14
New Data: CA COVID-19 Hospital Data


elaguerta:
County or Counties: All
Data source: https://data.chhs.ca.gov/dataset/california-covid-19-hospital-data-and-case-statistics/resource/6cd8d424-dfaa-4bdd-9410-a3d656e1176e?inner_span=True
Output to:
Update frequency/time: Daily, likely in the morning, updated to present day

benghancock:
Hi @elaguerta - I'd love to help with this issue, if you're still looking for help on it. It looks like the CSV data is pretty tidy, and as noted, pulls the hospital data for all counties. The columns are:
```
['county', 'todays_date', 'hospitalized_covid_confirmed_patients',
       'hospitalized_suspected_covid_patients', 'hospitalized_covid_patients',
       'all_hospital_beds', 'icu_covid_confirmed_patients',
       'icu_suspected_covid_patients', 'icu_available_beds']
``` 
How do you envision this integrating with the county scrapers? Would each county scraper also pull from this dataset, and filter for the selected county? Or would it not be integrated -- i.e. the hospital data would be scraped separately?

elaguerta:
Hi @benghancock, for now I think we should keep it separate because it is coming from a completely different source than all the county sources. Any integration that might be needed by the front-end design, or by our future API users, can happen downstream. To make that easier, I'd suggest you structure your top method to get just one county by a name or other identifier passed as an argument.

benghancock:
Got it, that makes sense. I'll plan to start working on this later today. I've forked the project and will create a branch for this issue. Thanks!

benghancock:
I've got the pagination working to loop through and collect all records returned by the API. Still need to collect the notes for each field, and document the output data model. By default the response returns all the records going back to March, which is a lot (5,000+ for all counties) and will obviously continue to grow. Do we want to limit the time period or just pull them all in so we have complete data each time?

benghancock:
Hi @elaguerta - I've put in a PR that hopefully delivers what's needed for this.

PR #57, opened 2020-05-17
Add Sonoma Data Scraper

ldtcooper:


ldtcooper:
The code is mostly done now. At this point, I'm just doing some clean-up and clarification, hence the draft status. 
I know my implementation is somewhat repetitive, and I'm open to suggestions to fix that. I played around with the idea of having a generic function to get the rows and cells from a tag, loop through them, and take a function to do any extra cleaning/transformation on them, but that made things more complicated and hard-to-read IMO.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+        transmissions[type] = int(number)
+    return transmissions
+
+def transform_tests(tests_tag: element.Tag) -> Dict[str, int]:
+    tests = {}
+    rows = get_rows(tests_tag)
+    for row in rows:
+        result, number, _pct = get_cells(row)
+        lower_res = result.lower()
+        tests[lower_res] = int(number.replace(',', ''))
```
Mr0grog, line 102: It feels like a helper for parsing integers might be useful:

```py
def parse_int(text: str) -> int:
    text = text.strip()
    if text == '-':
        return 0
    else:
        return int(text.replace(',', ''))
```

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    Transform function for tables which don't require any special processing.
+    Takes in a BeautifulSoup tag for a table and returns a dictionary
+    in which the keys are strings and the values integers
+    """
+    categories = {}
+    rows = get_rows(tag)
+    for row in rows:
+        cat, cases, _pct = get_cells(row)
+        categories[cat] = int(cases)
```
Mr0grog, line 115: Hopefully it’s never an issue, but I’ll bet you’re going to need similar parsing logic as above with commas if this ever grows big enough.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    return categories
+
+def get_unknown_race(race_eth_tag: element.Tag) -> int:
+    """
+    Gets the notes under the 'Cases by race and ethnicity' table to find the
+    number of cases where the person's race is unknown
```
Mr0grog, line 121: Oh, Sonoma, you could have just put it in the table. 😐

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    race_cases = {}
+    race_transform = {'Asian/Pacific Islander, non-Hispanic': 'Asian', 'Hispanic/Latino': 'Latinx_or_Hispanic', 'Other*, non-Hispanic': 'Other', 'White, non-Hispanic': 'White'}
+    rows = get_rows(race_eth_tag)
+    for row in rows:
+        group_name, cases, _pct = get_cells(row)
+        if group_name not in race_transform:
+            raise FutureWarning('The racial group {0} is new in the data -- please adjust the scraper accordingly')
+        internal_name = race_transform[group_name]
+        race_cases[internal_name] = int(cases)
```
Mr0grog, line 145: Same here with int parsing — this will probably break with values > 999.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    return(int(matches.groups()[0]))
+
+def transform_race_eth(race_eth_tag: element.Tag) -> Dict[str, int]:
+    """
+    Takes in the BeautifulSoup tag for the cases by race/ethnicity table and
+    transforms it into an object of form:
+    'race_eth': {'Asian': -1, 'Latinx_or_Hispanic': -1, 'Other': -1, 'White':-1, 'Unknown': -1}
+    NB: These are the only races reported seperatley by Sonoma county at this time
+    """
+    race_cases = {}
```
Mr0grog, line 137: In Alameda, I believe @elaguerta loaded the defaults from the data model JSON so that any race/ethnicities we don’t have values for come through as `-1` or `0` instead of being unlisted. We should *probably* do the same here.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    Takes in a BeautifulSoup tag of the cases by hospitalization table and
+    returns a dictionary with the numbers of hospitalized and non-hospitalized
+    cases
+    """
+    hospitalizations = {}
+    rows = get_rows(hospital_tag)
+    for row in rows:
+        hospitalized, number, _pct = get_cells(row)
+        if hospitalized == 'Yes':
```
Mr0grog, line 159: Maybe worth a case-insensitive comparison, just in case this changes in the future.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    Takes in a BeautifulSoup tag representing the percent of cases hospitalized
+    by gender and returns a dictionary of those percentages in float form
+    e.g. 9% is 0.09
+    """
+    hospitalized = {}
+    rows = get_rows(hospital_tag)
+    for row in rows:
+        gender, no, yes = get_cells(row)
+        yes_int = int(yes.replace('%', ''))
+        hospitalized[gender] = (yes_int / 100)
```
Mr0grog, line 176: Would it be useful to lower-case gender here?

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    return hospitalized
+
+def get_county() -> Dict:
+    """Main method for populating county data .json"""
+    url = 'https://socoemergency.org/emergency/novel-coronavirus/coronavirus-cases/'
+    page = requests.get(url)
```
Mr0grog, line 182: Might be useful to call `page.raise_for_status()` just in case you got an error page.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    sonoma_soup = BeautifulSoup(page.content, 'html.parser')
```
Mr0grog, line 183: Might be good to use `html5lib` instead of `html.parser`.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    tables = sonoma_soup.findAll('table')[4:] # we don't need the first three tables
+
+    try:
+        # we have a lot more data here than we are using
+        hist_cases, cases_by_source, cases_by_race, total_tests, cases_by_region, region_guide, hospitalized, underlying_cond, symptoms, cases_by_gender, underlying_cond_by_gender, hospitalized_by_gender, symptoms_female, symptoms_male, symptoms_desc, cases_by_age, symptoms_by_age, underlying_cond_by_age = tables
```
Mr0grog, line 188: This is super convenient, but I worry about the order changing here. Maybe safer to call `sonoma_soup.find(re.compile(r'^h\d$'), string='whatever')` for each?

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    model = {
+        'name': 'Sonoma County',
+        'update_time': generate_update_time(sonoma_soup),
+        'source': url,
+        'meta_from_source': get_source_meta(sonoma_soup),
+        'meta_from_baypd': 'Racial "Other" category includes "Black/African American, American Indian/Alaska Native, and Other"',
+        'series': transform_cases(hist_cases),
+        'case_totals': {
+            'transmission_cat': transform_transmission(cases_by_source),
+            'age_group': generic_transform(cases_by_age),
```
Mr0grog, line 201: This gets us the wrong format:

```js
        "age_group": {
            "0-17": 58,
            "18-49": 190,
            "50-64": 78,
            "65 and Above": 47,
            "Under Investigation": 0
        },
```

But we should have a list, and the group names should be slightly different: https://github.com/sfbrigade/data-covid19-sfbayarea/tree/master/data_models#cases-2

```js
[
	{"group": "0_to_17", "raw_count": -1 },
	{"group": "18_to_49", "raw_count": -1 },
	{"group": "50_to_64", "raw_count": -1 },
	{"group": "65_and_older", "raw_count": -1 },
	{"group": "Unknown", "raw_count": -1 }
]
```

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+            'race_eth': transform_race_eth(cases_by_race),
+            'gender': generic_transform(cases_by_gender)
```
Mr0grog, line 203: This is outputting different keys than we have for hospitalizations below. We should unify them.

Here it’s plural:

```js
        "gender": {
            "Males": 178,
            "Females": 195
        }
```

In hospitalizations it’s singular:

```js
    "hospitalizations": {
        "hospitalized_cases": {
            "hospitalized": 35,
            "not_hospitalized": 338
        },
        "gender": {
            "Male": 0.08,
            "Female": 0.1
        }
    }
```

data_scrapers/sonoma_county.py
elaguerta: Our data_model.json had inconsistent use of "age_group" values (only 1 instance was a list). I changed all the "age_group" tables to be lists in [#52, in progress](https://github.com/sfbrigade/data-covid19-sfbayarea/pull/52). 
As far as using the group names, I realized I need to go do that for Alameda County and SF county. And we will probably have to manually order the list by group-name, since they're often coming through as unordered .json objects (in curly braces).

data_scrapers/sonoma_county.py
elaguerta: I thought we would use the [chhs](https://data.chhs.ca.gov/dataset/california-covid-19-hospital-data-and-case-statistics/resource/6cd8d424-dfaa-4bdd-9410-a3d656e1176e?inner_span=True) as a common source for all hospitalization records. It looks like this is where most counties are getting their information, and we can be sure to have some sort of standard across counties by going straight to the chhs API. This referenced in issue #29. This is why the data_model doesn't currently include a hospitalization sub-table, I thought we could add it separately from the individual county scrapers. I decided to exclude hospitalization data available through SF county, to avoid confusion when we do get to #29.

README.md
```diff
@@ -9,4 +9,158 @@ To install this project, you can simply run `sh install.sh` in your terminal. Th
 To run the scraper, you can use the run script by typing `sh run_scraper.sh` into your terminal. This will enable the virtual environment and run `scraper.py`. Once again, the virtual environment will not stay active after the script finishes running. If you want to run the scraper without the run script, enable the virtual environment, then run `python3 scraper.py`.
 
 ## Running the API
-The best way to run the API right now is to run the command `FLASK_APP="app.py" FLASK_ENV=development flask run;`. Note that this is not the best way to run the scraper at this time.
\ No newline at end of file
+The best way to run the API right now is to run the command `FLASK_APP="app.py" FLASK_ENV=development flask run;`. Note that this is not the best way to run the scraper at this time.
+
+## Data Model
```
elaguerta, line 14: Do you want to merge the .data_scrapers/README  into the top level README? See PR #48

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+        new_deaths = dead - cumul_deaths
+        deaths.append({ 'date': date, 'deaths': new_deaths, 'cumul_deaths': dead })
+
+        new_recovered = recoveries - cumul_recovered
+        recovered.append({ 'date': date, 'recovered': new_recovered, 'cumul_recovered': recoveries })
+
+        new_active = active_cases - cumul_active
+        active.append({ 'date': date, 'active': new_active, 'cumul_active': active_cases })
+
+    return { 'cases': cases, 'deaths': deaths, 'recovered': recovered, 'active': active }
```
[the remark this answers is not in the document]
elaguerta: Do we want to include "active" and "recovered" time-series even if they're not in the data model? I believe Sonoma is the only county with these.  I'm personally suspicious of the label "recovered" when things like strokes are affecting people long after they test as negative. We're also excluding all the detailed breakdowns Sonoma has over symptoms and underlying conditions, which makes me ask the question: if data isn't in our data model, are there criteria we're applying to decide whether to include or exclude it?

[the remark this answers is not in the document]
ldtcooper: I put active and recovered in because they were there and I'd rather err on the side of putting too much data in. If we don't want to use them, then I'll take them out of the output.

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    cumul_deaths = 0
+    recovered = []
+    cumul_recovered = 0
+    active = []
+    cumul_active = 0
+    rows = get_rows(cases_tag)
+    for row in rows:
+        row_cells = row.findAll(['th', 'td'])
+        # print(type(row_cells))
+        date = row_cells[0].text.replace('/', '-')
```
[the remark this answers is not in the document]
ldtcooper: All of the dates are now using `dateutils`

data_scrapers/sonoma_county.py
```diff
@@ -0,0 +1,216 @@
...
+    {'community': -1, 'from_contact': -1, 'travel': -1, 'unknown': -1}
+    """
+    transmissions = {}
+    rows = get_rows(transmission_tag)
+    # turns the transmission categories on the page into the ones we're using
+    transmission_type_conversion = {'Community': 'community', 'Close Contact': 'from_contact', 'Travel': 'travel', 'Under Investigation': 'unknown'}
+    for row in rows:
+        type, number, _pct = get_cells(row)
+        if type not in transmission_type_conversion:
+            raise FutureWarning('The transmission type {0} was not found in transmission_type_conversion'.format(type))
```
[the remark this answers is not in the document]
ldtcooper: I created a custom `FormatError` exception we can use throughout the project

data_scrapers/sonoma_county.py
ldtcooper: 🤷

ldtcooper:
Alright @Mr0grog I think I'm ready for a second round of review

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,299 @@
...
+    for row in rows:
+        gender, cases, _pct = get_cells(row)
+        if gender not in gender_string_conversions:
+            raise FormatError('An unrecognized gender has been added to the gender table')
+        categories[gender_string_conversions[gender]] = parse_int(cases)
+    return categories
+
+# not sure why I need object in here: I've checked the output value types and
+# they are strings and ints, but mypy is convinced there is an object in there
+def age_transform(tag: element.Tag) -> List[Dict[str, Union[str, int, object]]]:
```
Mr0grog, line 159: Minor nit: should use the same naming convention as the other transform functions, which are all `def transform_xyz`.

> not sure why I need object in here: I've checked the output value types and they are strings and ints, but mypy is convinced there is an object in there

Oh, this is super interesting! I think it has to do with [how Mypy uses contextual info](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#context-in-type-inference). For example, this is OK:

```py
def return_union() -> Dict[str, Union[str, int]]:
    a = 'hi'
    b = 5
    return {'a': a, 'b': b}
```

But this causes the problem you’re seeing:

```py
def return_union() -> Dict[str, Union[str, int]]:
    a = 'hi'
    b = 5
    result = {'a': a, 'b': b}
    return result
```

Based on the above docs, I’m all but certain what’s happening is that when Mypy sees:

```py
{'a': a, 'b': b}
# Or in our case, {'group': age_string_transform[group], 'raw_cases': raw_cases}
```

…it has to infer the correct value type. There are a *lot* of valid ways to do that, and what it goes for is the most common super-type of all the values, which is `object` (both `int` and `str` are subtypes of `object`, as is basically everything — it’s pretty similar to `Any`). So the type Mypy infers is `Dict[str, object]`.

BUT when it sees:

```py
return {'a': a, 'b': b}
```

It’s contextually aware enough to know that, because this is being returned, it should go look at the declared return value and see if that would *also* be a compatible alternative. It is, so Mypy just infers that the type of the expression is the one you declared (`Dict[str, Union[str, int]]`).

In the former case, where there wasn’t an immediate return in front of the dict expression, Mypy has no way to tell that it should compare it with the declared return type to see if that’s a better alternative. Then by the time it gets to a return statement, all it knows is that all the dicts you put in the `categories` list are of type `Dict[str, object]`.

**Anyway, long story short, there are two options that can fix this:**

1. When you declare `categories`, declare its type, and append the dict expression to it instead of a variable with the dict (it’s looking at the context again):

    ```py
    categories: List[Dict[str, Union[str, int]]] = []
    ...
        categories.append({'group': age_string_transform[group],
                           'raw_cases': raw_cases})
    ```

2. Declare the dict type before appending it down on line 181:

    ```py
    # Give Mypy a better type to guess than Dict[str, object]
    element: Dict[str, Union[str, int]] = {
        'group': age_string_transform[group],
        'raw_cases': raw_cases
    }
    categories.append(element)
    ```

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,299 @@
...
+    returns a dictionary with the numbers of hospitalized and non-hospitalized
+    cases
+    """
+    hospitalizations = {}
+    rows = get_rows(hospital_tag)
+    for row in rows:
+        hospitalized, number, _pct = get_cells(row)
+        if hospitalized.lower() == 'yes':
+            hospitalizations['hospitalized'] = parse_int(number)
+        else:
```
Mr0grog, line 238: Might be worth checking the value here anyway. (And also raising an error is `hospitalized.lower()` is something totally unexpected.) What if the row titles change to `Hospitalized` and `Not Hospitalized`? (Or something else.) Then we’d wind up with:

```js
{
  // Note "hospitalized" is not even present.
  "not_hospitalized": <value of last row in table, regardless of what it represents>
}
```

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,299 @@
...
+
+def get_county() -> Dict:
+    """
+    Main method for populating county data .json
+    """
+    url = 'https://socoemergency.org/emergency/novel-coronavirus/coronavirus-cases/'
+    page = requests.get(url)
+    page.raise_for_status()
+    sonoma_soup = BeautifulSoup(page.content, 'html5lib')
+    tables = sonoma_soup.find_all('table')[4:] # we don't need the first three tables
```
Mr0grog, line 271: Looks like this line is vestigial.

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,299 @@
...
+    cumul_cases = 0
+    deaths = []
+    cumul_deaths = 0
+    # recovered = []
+    # cumul_recovered = 0
+    # active = []
+    # cumul_active = 0
+    rows = get_rows(cases_tag)
+    for row in rows:
+        row_cells = row.find_all(['th', 'td'])
```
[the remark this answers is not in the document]
Mr0grog: To be clear, it’s cool if the answer is “tried it, but too much of a pain/not worth the tradeoff.” I just want to make sure it didn’t get missed entirely since I’m concerned about changing columns.

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,299 @@
...
+        raise ValueError(f'Date is not in ISO 8601'
+                         f'format: "{update_time_text}"')
+    return date.isoformat()
+
+def get_source_meta(soup: BeautifulSoup) -> str:
+    """
+    Finds the 'Definitions' header on the page and gets all of the text in it
+    """
+    definitions_header = soup.find('h3', string='Definitions')
+    definitions_text = definitions_header.find_parent().text
```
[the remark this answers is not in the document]
ldtcooper: Yes, good catch!

data_scrapers/sonoma.py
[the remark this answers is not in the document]
ldtcooper: Oh, sorry, I missed that suggestion completely. I'll give it a try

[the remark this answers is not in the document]
ldtcooper: I started trying this, and I think it's going to be more work than is worth it right now. The shape of the tables has been consistent since I started working on this, although the order of the tables on the page has changed. The function to parse tables is in there, and we can switch over to using it if the table format starts changing

data_scrapers/sonoma.py
```diff
@@ -38,6 +38,22 @@ def get_cells(row: element.ResultSet) -> List[str]:
     """
     return [el.text for el in row.find_all(['th', 'td'])]
 
+def row_list_to_dict(row_list: List[str], headers: List[str]) -> Dict[str, str]:
+    output = {}
+    for i in range(len(row_list)):
+        val = row_list[i]
+        header = headers[i]
+        output[header] = val
+    return output
```
Mr0grog, line 47: This whole function can actually just be done with:

```py
dict(zip(headers, row_list))
```

So you could simplify this function to:

```py
def row_list_to_dict(row_list, headers):
    return dict(zip(headers, row_list))
```

Or get rid of it entirely and replace line 55 with:

```py
return [dict(zip(header_cells, row)) for row in body_cells]
```

data_scrapers/sonoma.py
```diff
@@ -38,6 +38,22 @@ def get_cells(row: element.ResultSet) -> List[str]:
...
+
+def parse_table(tag: element.Tag) -> List[Dict[str, str]]:
+    rows = tag.find_all('tr')
+    header = rows[0]
+    body = rows[1:]
+    header_cells = get_cells(header)
+    body_cells = [get_cells(row) for row in body]
```
Mr0grog, line 54: It’s a minor performance improvement, but you could make this a generator comprehension:

```suggestion
    body_cells = (get_cells(row) for row in body)
```

Less memory (no extra list is created for `body_cells`) and fewer iterations, so less CPU, too.

data_scrapers/sonoma.py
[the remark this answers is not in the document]
Mr0grog: OK.

> we can switch over to using it if the table format starts changing

But just to be clear, part of what I was getting at is that it might be hard to notice that the format has started changing. With code like this:

```py
def transform_cases(...):
    ...
    active_cases, new_infected, dead, recoveries = [parse_int(el.text) for el in row_cells[1:]]
    ....
```

If those columns continue to be numeric, nothing about this program will tell you that you’re getting the wrong values out when things were rearranged, new columns were inserted, etc.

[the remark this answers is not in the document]
Mr0grog: (This would also help alert you to the fact that the order of tables has changed, since the re-ordered tables would presumably have different headers than you expected, causing errors when you try to get cells by their header.)

data_scrapers/sonoma.py
```diff
@@ -0,0 +1,324 @@
...
+def transform_gender_hospitalizations(hospital_tag: element.Tag) -> Dict[str, float]:
+    """
+    Takes in a BeautifulSoup tag representing the percent of cases hospitalized
+    by gender and returns a dictionary of those percentages in float form
+    e.g. 9% is 0.09
+    """
+    hospitalized = {}
+    rows = get_rows(hospital_tag)
+    for row in rows:
+        gender, no, yes = get_cells(row)
```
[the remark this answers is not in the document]
ldtcooper: I ended up taking out the hospitalization code. I want to start looking into refactoring it, but I'd rather get a working version out before that.

README.md
Mr0grog: @ldtcooper any thoughts on this point from @elaguerta? There are a lot of great improvements in #48 and it seems like we should have merged it quite a while ago.

This is kind of in conflict with that. I think we should merge #48. I’m not sure whether you’re thinking we should also move all that content to the top level README (seems pretty big and maybe a bit unwieldy, but that is not a strongly held opinion).

elaguerta:
Heads up, I also just merged #48 so that may conflict with your `README.md` s

ldtcooper:
It looks like it does. I'll try to fix those conflicts and linter errors tonight

ldtcooper:
Looks like all the linter and merge problems have been fixed @Mr0grog @elaguerta

README.md
Mr0grog: So this now repeats a lot of information that is already in [`data_models/README.md`](https://github.com/sfbrigade/data-covid19-sfbayarea/tree/sonoma/data_models#age-group). Would it be better to just link to it? That way we don’t get in a situation where someone updates one place but not the other.

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,273 @@
+#!/usr/bin/env python3
+import requests
+import json
+import re
+import dateutil.parser
+from typing import List, Dict, Union
+from bs4 import BeautifulSoup, element # type: ignore
+from format_error import FormatError # type: ignore
```
Mr0grog, line 8: This is broken for me. If I run `./run_scraper_data.sh solano` or `python scraper_data.py solano`, I get:

```
  File "scraper_data.py", line 4, in <module>
    from covid19_sfbayarea import data as data_scrapers
  File "/Users/rbrackett/Dev/sfbrigade/data-covid19-sfbayarea/covid19_sfbayarea/data/__init__.py", line 4, in <module>
    from . import sonoma
  File "/Users/rbrackett/Dev/sfbrigade/data-covid19-sfbayarea/covid19_sfbayarea/data/sonoma.py", line 8, in <module>
    from format_error import FormatError
ModuleNotFoundError: No module named 'format_error'
```

Does it work for you? Seems like it should be:

```suggestion
from .format_error import FormatError
```

(It also shouldn’t need need the `# type: ignore` comment.)

---

Also, while you’re here, maybe it’s worthwhile to make this a shared module up in `covid19_sfbayarea/errors.py` instead of `covid19_sfbayarea/data/format_error.py`? Some of the news scrapers also use [an identical class](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/47ced984c3379fa4b5908e3de1e9778714d5e7c4/covid19_sfbayarea/news/errors.py#L1-L7), and we should really only have one implementation. Then this line would be:

```suggestion
from ..errors import FormatError
```

README.md
ldtcooper: I thought we were merging `data_models/README.md` into the main readme

Mr0grog: OK! I was unclear on exactly which direction we were going there. If we’re merging into the top-level README, then this is missing a bunch of stuff from `data_models/README.md`, and this PR should probably also remove that file.

elaguerta:
They are using ArcGIS dashboards now. You can look at Alameda as a model. 
I found some endpoints.

- **Cases and deaths reported by Date:** https://services1.arcgis.com/P5Mv5GY5S66M8Z1Q/arcgis/rest/services/NCOV_Cases_Sonoma_County/FeatureServer
- **Hodge Podge of Tables Appended Together, including Age Group, Gender, Test Results, and Race Ethnicity:** https://services1.arcgis.com/P5Mv5GY5S66M8Z1Q/ArcGIS/rest/services/NCOV_Cases_Sonoma_County_Statistics/FeatureServer

You might find this post helpful: https://stackoverflow.com/questions/50161492/how-do-i-scrape-data-from-an-arcgis-online-map

Also, we may want to address #85 now that there are 2 counties using ArcGIS.

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    race_cases = {
+        'African_Amer': 0,
+        'Asian': 0,
+        'Latinx_or_Hispanic': 0,
+        'Native_Amer':0,
+        'Multiple_Race':0,
+        'Other': 0,
+        'Pacific_Islander': 0,
+        'White': 0,
+        'Unknown': 0
```
Mr0grog, line 207: Seems like the base value for `African_Amer`, `Native_Amer`, `Multiple_Race`, and `Pacific_Islander` should probably be either `-1` or not present here to clarify that the counts are unknown rather than 0 (since they’re included in `Asian` or `Other`).

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    Takes in a BeautifulSoup tag for a table and returns a list of
+    dictionaries in which the keys are strings and the values integers
+    """
+    categories: List[Dict[str, Union[str, int]]] = []
+    rows = get_rows(tag)
+    for row in rows:
+        group, cases, *rest = get_cells(row)
+        raw_cases = parse_int(cases)
+
+        element: Dict[str, Union[str, int]] = {'group': group, 'raw_cases': raw_cases}
```
Mr0grog, line 175: This should be `raw_count`:

```suggestion
        element: Dict[str, Union[str, int]] = {'group': group, 'raw_count': raw_cases}
```

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    cumul_cases = 0
+    deaths = []
+    cumul_deaths = 0
+    # recovered = []
+    # cumul_recovered = 0
+    # active = []
+    # cumul_active = 0
+    rows = reversed(get_rows(cases_tag))
+    for row in rows:
+        row_cells = row.find_all(['th', 'td'])
```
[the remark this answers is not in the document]
ldtcooper: I think there was, but it seems to be working using `get_cells` so I changed it

README.md
Mr0grog: So… we’re just going to put this in here as-is?

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+def row_list_to_dict(row: List[str], headers: List[str]) -> Dict[str, str]:
+    return dict(zip(headers, row))
+
+def parse_table(tag: element.Tag) -> List[Dict[str, str]]:
+    rows = tag.find_all('tr')
+    header = rows[0]
+    body = rows[1:]
+    header_cells = get_cells(header)
+    body_cells = (get_cells(row) for row in body)
+    return [row_list_to_dict(row, header_cells) for row in body_cells]
```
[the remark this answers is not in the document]
Mr0grog: Since you asked for a re-review, does this mean you’re going to keep these?

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+    Generates a timestamp string (e.g. May 6, 2020 10:00 AM) for when the scraper is run
+    """
+    update_time_text = soup.find('time', {'class': 'updated'}).text.strip()
+    try:
+        date = dateutil.parser.parse(update_time_text)
+    except ValueError:
+        raise ValueError(f'Date is not in ISO 8601'
+                         f'format: "{update_time_text}"')
+    return date.isoformat()
```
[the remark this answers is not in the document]
Mr0grog: Keeping this without timezone info?

ldtcooper:
I did just miss those. Let me take a look

covid19_sfbayarea/data/sonoma.py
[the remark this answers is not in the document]
ldtcooper: I ended up redoing all the transform functions to use the new functions

covid19_sfbayarea/data/sonoma.py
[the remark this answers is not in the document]
ldtcooper: I went with the `datetime` solution to get this working

ldtcooper:
Okay, I think I addressed those three points

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    # this lets us get the second cases table
+    return tables[-1]
+
+def get_cells(row: element.ResultSet) -> List[str]:
+    """
+    Gets all th and tr elements within a single tr element
+    """
+    return [el.text for el in row.find_all(['th', 'td'])]
+
+def row_list_to_dict(row: List[str], headers: List[str]) -> UnformattedSeriesItem:
```
Mr0grog, line 39: Minor nit: one of the important things about naming (or aliasing) types like this is to change how you conceptualize your values and functions (e.g. you shouldn’t be thinking of `UnformattedSeriesItem` like a shortcut for `Dict[str, str]` here; you should be thinking of it like a subclass of `dict` — it should conceptually be its own separate thing).

So if you’re changing the return type to something named `UnformattedSeriesItem`, it’s probably a good idea to change the function name to not talk about making a `dict` and instead call it something like `row_list_to_series_item` or something.

covid19_sfbayarea/data/sonoma.py
```diff
@@ -0,0 +1,261 @@
...
+    Where each list contains dictionaries (representing each day's data)
+    of form (example for cases):
+    { 'date': '', 'cases': -1, 'cumul_cases': -1 }
+    """
+    cases = []
+    cumul_cases = 0
+    deaths = []
+    cumul_deaths = 0
+
+    rows = list(reversed(parse_table(cases_tag)))
```
Mr0grog, line 103: Oh, forgot to comment on this in the review: I don’t think there’s any reason to convert this to a list, since you’re only iterating over it once and not returning it:

```suggestion
    rows = reversed(parse_table(cases_tag))
```

PR #58, opened 2020-05-21
Add more strict configuration for Mypy

Mr0grog:
Per a discussion in Slack, we should be a little bit more strict about requiring and checking type annotations. This adds a config file for Mypy and specifies a configuration that is more strict. It should help us be a little better about our types.

This required some updates to existing files, and will probably require changes to any in-progress PRs, so those will need to be rebased or updated once this lands.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,9 +1,11 @@
 from flask import Flask
 from scraper import pipeline
+from typing import Dict
+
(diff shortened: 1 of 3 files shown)


Mr0grog:
(I just fixed the breakage I made for the demo.)

PR #68, opened 2020-05-27
Scrape Alameda News

Mr0grog:
You can now get a news feed for Alameda county with:

```console
> python scraper_news.py alameda
```

This also fixes an issue with the default value for format. (You had to explicitly specify a format; the default was getting misinterpreted.)

### DIFF
--- a/news/alameda.py
+++ b/news/alameda.py
@@ -1,7 +1,6 @@
 from bs4 import BeautifulSoup  # type: ignore
 import dateutil.parser
 import dateutil.tz
-from selenium import webdriver  # type: ignore
 from typing import List
 from urllib.parse import urljoin
 from .base import NewsScraper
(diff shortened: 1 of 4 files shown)


Mr0grog:
@ldtcooper this should be good to go if it looks good to you. Then I’ll work with @kengoy to get this running with the frontend.

Mr0grog:
(Note: I had to add the Selenium stuff back in -- I’m guessing it’s doing something that is approving an IP address, which is why it started working at some point for me. Dunno exactly. 🤷)

news/alameda.py
```diff
@@ -1,6 +1,103 @@
...
+                date = dateutil.parser.parse(date_string)
+                if date.tzinfo is None:
+                    date = date.replace(tzinfo=PACIFIC_TIME)
+            except ValueError:
+                raise ValueError(f'Article {index} date could not be parsed: '
+                                 f'"{date_string}"')
+
+            # Some info cells just have the title as a link, while others have
+            # the title as text followed by links for each language.
+            english_link = info_cell.find('a', string='English')
```
ldtcooper, line 79: Is this a link for news in English as opposed to other languages?

Mr0grog: Ah, yeah. I should clarify that this case only happens when there are multiple languages for an item, and that’s rare (I think only two or three out of probably 15+ news items).

I figured the best approach (for now, at least) would be to just choose the English one in this case, since all the single-language items are English-only.

PR #69, opened 2020-05-28
Re-organize data and news scrapers under one top-level package

Mr0grog:
Instead of having two Python packages in this repo (`data_scrapers` and `news`), this PR puts them together as sub-packages under `covid19_sfbayarea`. The top-level package name is based on this repo. **Let me know if we should change it.**

This way, the two can share code. I've changed as little code as possible here -- mostly just made a lot of absolute imports into relative imports. The two top-level scripts are still the same so the frontend does have to change how it uses this repo at all. The `data_models` directory is still at the top level since it's not really Python code.

I’ve also added a shared module for getting a nicely configured version of Firefox webdriver. Whenever we create an instance of Selenium Firefox, we'll now get a headless one by default (i.e. it's invisible and doesn't require a fake display when running on a server). To get a normal, non-headless browser, set the `FIREFOX_VISIBLE` environment variable, e.g:

    > FIREFOX_VISIBLE=true python scraper_data.py alameda


Mr0grog:
@ldtcooper I’m thinking the `FormatError` class you’re adding in #57 could go into a shared `covid19_sfbayarea.errors` module kind of like how I handled the Firefox stuff here. https://github.com/sfbrigade/data-covid19-sfbayarea/blob/f1c7f0568f3379145977d1662cd9e7ac7760024f/data_scrapers/format_error.py#L1-L7

scraper_data.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import click
 import json
-import data_scrapers
+from covid19_sfbayarea import data as data_scrapers
```
Mr0grog, line 4: I was trying to keep names from getting too painfully long/nested, but I may have made this one too generic. Happy to change if anyone has better ideas than `data`.

Mr0grog:
Rebased to handle #70 getting merged.

Ping @ldtcooper @elaguerta @kwonangela7 this needs a quick review so it can get merged and other PRs can take advantage of it.

scraper_data.py
elaguerta: `data` is clear enough to me. I thought about `county_data` but we will be plugging into sources other than county health departments.

Mr0grog:
Rebased on master now that #72 has merged.

Mr0grog:
Rebased to include all the news scrapers already merged so far.

PR #73, opened 2020-06-02
Add news scraper for Solano County

### LINKED ISSUE #61: Add news scraper for Solano county
News scrapers live in the [`news` directory](https://github.com/sfbrigade/data-covid19-sfbayarea/tree/master/news). You can follow the [San Francisco scraper](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/master/news/san_francisco.py) as an example.

## Solano County

1. The [county homepage](http://www.solanocounty.com/) has a “headline news” section that we could scrape and filter for covid-related key terms like the Alameda County RSS feed above.

2. Has a [coronavirus-specific press releases page](http://www.solanocounty.com/depts/ph/coronavirus_links/coronavirus_press_releases_and_information.asp) we could scrape. No RSS.

Mr0grog:
This depends on #72 and fixes #61. The only files with changes from #70 are `news/__init__.py` and `news/solano.py`.

### DIFF
--- a/covid19_sfbayarea/news/__init__.py
+++ b/covid19_sfbayarea/news/__init__.py
@@ -7,7 +7,7 @@ from .napa import NapaNews
 from .san_francisco import SanFranciscoNews
 from .san_mateo import SanMateoNews
 from .santa_clara import SantaClaraNews
-# from .solano import SolanoNews
+from .solano import SolanoNews
 # from .sonoma import SonomaNews
 
 
@@ -19,6 +19,6 @@ scrapers: Dict[str, Type[NewsScraper]] = {
     'san_francisco': SanFranciscoNews,
     'san_mateo': SanMateoNews,
     'santa_clara': SantaClaraNews,
-    # 'solano': SolanoNews,
+    'solano': SolanoNews,
     # 'sonoma': SonomaNews,
 }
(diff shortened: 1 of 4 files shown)


Mr0grog:
Rebased on master now that #72 has merged.

news/solano.py
```diff
@@ -0,0 +1,92 @@
...
+        return NewsItem(id=url, url=url, title=title, date_published=date,
+                        summary=summary)
+
+    def is_covid_related(self, item: NewsItem) -> bool:
+        comparable = ' '.join([
+            item.title,
+            item.summary or '',
+            item.url
+        ]).lower()
+        terms = ('covid', 'coronavirus', 'health')
```
ldtcooper, line 91: I'm noticing some variation in what terms you're looking for to determine whether or not a news item is COVID-related. Is there a reason that Solano's scraper doesn't use (for example) "shelter in place" while Napa's does? Is there a reason to not use the same words for all counties?

Mr0grog: Ah, yeah, I really should have extracted this out into a utility function or a method on the `NewsScraper` base class. It basically just evolved as I continued working on the counties (so you can see that Sonoma is the newest because it’s the most complex).

I’ll pull this out so we only have one.

Mr0grog: Done. ✅

Mr0grog:
Ready for re-review.

PR #80, opened 2020-06-10
Add Marin County Scraper

kwonangela7:
- Please check in-line comments - specifically around the testing data I used.
- Er, I shouldn't have made any edits to the data_model.json file. I guess I added spaces lol. I'll change that...

I'll work on passing CircleCI tests, and will take another look through my code to see if there's anything else I wanted to point out, but I think for now I've highlighted the most notable comments.


covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+def get_case_series(chart_id: str, url: str) -> List:
+	"""This method extracts the date, number of cumulative cases, and new cases."""
+	csv_ = extract_csvs(chart_id, url)
+	series = []
+
+	csv_strs = csv_.splitlines()
+	keys = csv_strs[0].split(',')
+	
+	if keys != ['Date', 'Total Cases', 'Total Recovered*', 'Total Hospitalized', 'Total Deaths']:
+		raise ValueError('The headers have changed')
```
Mr0grog, line 109: If you use the build-in `csv.DictReader` class, you might not need this, since you’ll get a `KeyError` if you try to grab a value by a column name that doesn’t exist.

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+
+	case_history = []
+
+	for row in csv_strs[1:]:
+	 	daily = {}
+	 	# Grab the date in the first column
+	 	date_time_obj = datetime.strptime(row.split(',')[0], '%m/%d/%Y')
+	 	daily["date"] = date_time_obj.isoformat()
```
Mr0grog, line 117: I think this is supposed to be a `date`, not a `datetime`, right?

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	 	# Collect the case totals in order to compute the change in cases per day 
+	 	case_history.append(int(row.split(',')[1]))
+	 	# Grab the cumulative number in the fifth column
+	 	daily["cumul_cases"] = int(row.split(',')[1])
+	 	series.append(daily)
+		
+	case_history_diff = np.diff(case_history)
```
Mr0grog, line 124: Numpy needs to be added to `requirements.txt` if you’re going to use it here. BUT if all we’re using it for is calculating the difference between rows, is it really worthwhile/really simplifying things?

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	 	daily["date"] = date_time_obj.isoformat()
+	 	# Collect the case totals in order to compute the change in cases per day 
+	 	case_history.append(int(row.split(',')[1]))
+	 	# Grab the cumulative number in the fifth column
+	 	daily["cumul_cases"] = int(row.split(',')[1])
+	 	series.append(daily)
+		
+	case_history_diff = np.diff(case_history) 
+	# there will be no calculated difference for the first day, so adding it in manually
+	case_history_diff = np.insert(case_history_diff, 0, 0)
```
Mr0grog, line 126: This should be the total cases from the first row rather than 0, shouldn’t it? And then each row needs to have the total cases from the first row added to it.

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	if keys != ['Age Category', 'POPULATION', 'Cases', 'Hospitalizations', 'Deaths']:
+		raise ValueError('The headers have changed')
+
+	ages = ['0-18', '19-34', '35-49', '50-64', '65+'] 
+	for row in csv_strs[1:]:
+		c_age = {}
+		d_age = {}
+		# Extracting the age group and the raw count (the 3rd and 5th columns, respectively) for both cases and deaths.
+		# Each new row has data for a different age group.
+		c_age["group"] = row.split(',')[0]
```
Mr0grog, line 181: These still need to get mapped to the group names from https://github.com/sfbrigade/data-covid19-sfbayarea/tree/master/data_models, right?

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	csv_ = extract_csvs(chart_id, url)
+	series = []
+
+	csv_strs = csv_.splitlines()
+	keys = csv_strs[0].split(',')
+	
+	test_history = []
+	
+	# Grab the dates, which are in the header
+	for entry in csv_strs[:1][0].split(',')[1:]:
```
Mr0grog, line 258: Could this just be:

```suggestion
	for entry in csv_strs[0].split(',')[1:]: 
```

But also, like above, I’d recommend using the `csv` module for parsing, so this should probably be more different anyway. ¯\\\_(ツ)\_/¯

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	# Grab the dates, which are in the header
+	for entry in csv_strs[:1][0].split(',')[1:]: 
+		# need to exclude very first item in the csv_strs[:1][0].split(',') list (which is the value 'Date')
+		daily = {}
+		date_time_obj = datetime.strptime(entry, '%m/%d/%Y')
+		daily["date"] = date_time_obj.isoformat()
+		series.append(daily)
+
+	# The slicing makes this if statement hard to look at... there must be a better way?
+	if csv_strs[1:2][0].split(',')[:1][0] != 'Positive Tests' and csv_strs[2:][0].split(',')[:1][0] != 'Negative Tests':
```
Mr0grog, line 266: If you’re just trying to get a single entry from a list, you can just do `list[index]`. e.g:

```py
csv_strs[1:2][0]
# Is the same as:
csv_strs[1]

# And:
csv_strs[1].split(',')[:1][0]
# Is the same as:
csv_strs[1].split(',')[0]
```

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+		raise ValueError('The kinds of tests have changed.')
+
+	# Grab the positive test result numbers, which is in the second row. 
+	# [1:] is included to make sure that 'Positive Tests' is not captured.
+	p_entries = csv_strs[1:2][0].split(',')[1:]
+	n_entries = csv_strs[2:][0].split(',')[1:]
+	
+	get_test_series_helper(series, p_entries, ['positive', 'cumul_pos'])
+	get_test_series_helper(series, n_entries, ['negative', 'cumul_neg'])
```
Mr0grog, line 275: To keep this whole thing simpler, it might be helpful to flip the columns and rows using [`zip`](https://docs.python.org/3.8/library/functions.html#zip):

```py
# Take several series (dates, positive cases, negative cases) and turn them
# into tuples of (date, positive, negative), one for each column.
dates, positives, negatives = [row.split(',')[:1] for row in csv_strs]
time_series = zip(dates, positives, negatives)

# Or more compact, though possibly tougher to read:
time_series = zip(*[row.split(',')[:1] for row in csv_strs])

# Results in:
[
  ('3/4/2020', 1, 1),
  ('3/5/2020', 0, 2),
  ('3/6/2020', 0, 9),
  ('3/7/2020', 0, 0),
  etc...
]
```

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+from typing import List, Dict, Tuple
+from selenium import webdriver
+from bs4 import BeautifulSoup
+from urllib.parse import unquote_plus
+from datetime import datetime
+import re
+
+from .utils import get_data_model
+
+def get_county_data() -> Dict:
```
Mr0grog, line 14: This needs to be named `get_county`!

```suggestion
def get_county() -> Dict:
```

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+		# get pos/neg test count
+		day[keys[0]] = int(curr)
+		# add that day's pos/neg test count to get cumulative number of positive tests
+		cumul += curr
+		day[keys[1]] = cumul 
+		index += 1
+	return series
+
+
+get_county_data()
```
Mr0grog, line 303: This needs to go, `get_county_data()` needs to be renamed and should return the dict instead of printing, and `covid19_sfbayarea/data/__init__.py` needs to list this module. :)

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/cf49d5a6d3efe0ac43f93dcd0573c5e4b38c4dde/covid19_sfbayarea/data/__init__.py#L2-L15

That’s what’s required so this can run the same way the other scrapers do:

```console
$ bash run_scraper_data.sh marin
# Or:
$ python scraper_data.py marin

# Or to dump all counties into a directory:
$ bash run_scraper_data.sh --output county_data
# Or:
$ python scraper_data.py --output county_data
```

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	"""Main method for populating county data"""
+
+	url = 'https://coronavirus.marinhhs.org/surveillance'
+	model = get_data_model()
+
+	chart_ids = {"cases": "Eq6Es", "deaths": "Eq6Es", "tests": '2Hgir', "age": "VOeBm", "gender": "FEciW", "race_eth": "aBeEd"} 
+	# population totals and transmission data missing.
+	model['name'] = "Marin County"
+	model['update_time'] = datetime.today().isoformat()
+	# No actual update time on their website? They update most charts daily (so the isoformat is only partially correct.)
```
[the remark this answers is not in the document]
kwonangela7: Not sure if it makes sense to use the date in the updated column. It's certain that the updated column is not representative of the whole page (when I check now, the updated column for those two charts is 6/19, but the last data point on the "Daily Count of Positive Results and Total Tests for Marin County Residents by Test Date" chart is 6/15). 

One way around this would be to note the last updated date for each set of data in the output (aka each csv), but then there's a problem b/c the race/ethnicity data don't have any indication of updated dates whatsoever...

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	model["case_totals"]["age_group"], model["death_totals"]["age_group"] = get_breakdown_age(chart_ids["age"], url)
+	model["case_totals"]["gender"], model["death_totals"]["gender"] = get_breakdown_gender(chart_ids["gender"], url)
+	model["case_totals"]["race_eth"], model["death_totals"]["race_eth"] = get_breakdown_race_eth(chart_ids["race_eth"], url)
+	
+	print(model)
+
+def extract_csvs(chart_id: str, url: str) -> str:
+	"""This method extracts the csv string from the data wrapper charts."""
+	driver = webdriver.Chrome('/Users/angelakwon/Downloads/chromedriver') 
+	# need to figure out how to change the webdriver
```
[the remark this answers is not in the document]
kwonangela7: I'll be using Firefox! It just didn't occur to me lol.

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	frame = driver.find_element_by_css_selector(f'iframe[src*="//datawrapper.dwcdn.net/{chart_id}/"]')
+
+	driver.switch_to.frame(frame)
+	# Grab the raw data out of the link's href attribute
+	csv_data = driver.find_element_by_class_name('dw-data-link').get_attribute('href')
+	# Switch back to the parent frame to "reset" the context
+	driver.switch_to.parent_frame()
+	
+	# Deal with the data
+	if csv_data.startswith('data:'):
```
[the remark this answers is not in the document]
kwonangela7: good callout. thanks!

covid19_sfbayarea/data/marin_scraper.py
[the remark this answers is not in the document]
Mr0grog: > It's certain that the updated column is not representative of the whole page

Then my idea definitely sounds like a bad one! 😅 

> An option is to note the last updated date for each element in the scraper_output["series"]. Since the the race/ethnicity/age data that doesn't have any indication of updated dates whatsoever, I could just say the last updated date is N/A or missing. How's that?

My 2¢: I think what you’re already doing here is OK (except it would still be good to note in `meta_from_baypd`), and am not sure it’s worth finding a way to note the update time for each element of the dataset — especially since we don’t have that information in an especially consistent way.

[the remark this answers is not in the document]
kwonangela7: Thanks for your feedback! I'll make sure to note it accordingly.

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+		if media != 'application/octet-stream;charset=utf-8':
+			raise ValueError(f'Cannot handle media type "{media}"')
+		csv_string = unquote_plus(data)
+
+	# Then leave the iframe
+	driver.switch_to.default_content()
+
+	return csv_string
+
+def get_metadata(url: str, chart_ids: str) -> Tuple:
```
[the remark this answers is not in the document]
kwonangela7: whoops, yes.

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+
+	to_be_matched = ['Total Cases, Recovered, Hospitalizations and Deaths by Date Reported', 'Daily Count of Positive Results and Total Tests for Marin County Residents by Test Date ', 'Cases, Hospitalizations, and Deaths by Age, Gender and Race/Ethnicity ']
+	chart_metadata = []
+
+	for text in to_be_matched:
+		target = soup.find('h4',text=text)
+		if not target:
+			raise ValueError('Cannot handle this header.')
+		for sib in target.find_next_siblings()[:1]: # I only want the first paragraph tag
+			# Is it more efficient to use something like (soup object).select('h1 + p') to grab the first paragraph that follows?
```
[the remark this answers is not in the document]
kwonangela7: :face-palm: yes, it definitely makes more sense to just do that and not use a for loop.

I'll revisit the source code and see if it makes sense to use soup.select instead. thanks!

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	# Deal with the data
+	if csv_data.startswith('data:'):
+		media, data = csv_data[5:].split(',', 1)
+		# Will likely always have this kind of data type
+		if media != 'application/octet-stream;charset=utf-8':
+			raise ValueError(f'Cannot handle media type "{media}"')
+		csv_string = unquote_plus(data)
+
+	# Then leave the iframe
+	driver.switch_to.default_content()
```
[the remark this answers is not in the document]
kwonangela7: i did some quick googling and I think they do slightly different things? I'll dig into it more and get rid of one if it seems that it's redundant. Thanks!

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: thanks Rob! I'll rename the file at the end when this PR is ready to go (if I rename the file to marin.py now then these comments will be on a deleted file. Sorry about the inconvenience.

kwonangela7: thank you Rob!!

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: whoops, yes! Thanks for catching that :)

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: yes, thanks for catching this Rob!!

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: that makes sense LOL. i don't know why i was writing such strange code... thank you!

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+		frame = driver.find_element_by_css_selector(f'iframe[src*="//datawrapper.dwcdn.net/{chart_id}/"]')
+		driver.switch_to.frame(frame)
+		# The metadata for the charts is located in elements with the class `dw-chart-notes' 
+		for c in driver.find_elements_by_class_name('dw-chart-notes'):
+			chart_metadata.append(c.text)
+
+		# Switch back to the parent frame to "reset" the context
+		driver.switch_to.parent_frame()
+
+	driver.quit()
```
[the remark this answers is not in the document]
kwonangela7: cool! i'm going to try rewriting some code as context managers after I address the rest of the comments and fix the failed CI tests...

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+
+	for text in to_be_matched:
+		target = soup.find('h4',text=text)
+		if not target:
+			raise ValueError('Cannot handle this header.')
+		for sib in target.find_next_siblings()[:1]: # I only want the first paragraph tag
+			# Is it more efficient to use something like (soup object).select('h1 + p') to grab the first paragraph that follows?
+			metadata += [sib.text]
+
+	# Metadata for each chart visualizing the data of the csv file I'll pull. There's probably a better way to organize this.
```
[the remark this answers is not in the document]
kwonangela7: thanks for the suggestion! I might be misunderstanding something, but why should get_chart_meta call into a function that will switch to the chart's frame? The metadata function will be grabbing content not from frames but from simpler soup objects.

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: Yes, that variable should be date_obj! Unless that's not what you're referring to?

covid19_sfbayarea/data/marin_scraper.py
```diff
@@ -0,0 +1,303 @@
...
+	# Return the metadata. I take the set of the chart_metadata since there are repeating metadata strings.
+	return metadata, list(set(chart_metadata)) 
+
+def get_case_series(chart_id: str, url: str) -> List:
+	"""This method extracts the date, number of cumulative cases, and new cases."""
+	csv_ = extract_csvs(chart_id, url)
+	series = []
+
+	csv_strs = csv_.splitlines()
+	keys = csv_strs[0].split(',')
```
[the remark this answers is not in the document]
kwonangela7: that makes sense, thank you!

covid19_sfbayarea/data/marin_scraper.py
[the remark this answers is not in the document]
Mr0grog: > The metadata function will be grabbing content not from frames but from simpler soup objects.

If that’s the case, great! You can disregard what I said. I was responding to the fact that in the current code, it switches frames:

```py
def get_metadata(url: str, chart_ids: str) -> Tuple:
    ...
    # Metadata for each chart visualizing the data of the csv file I'll pull. There's probably a better way to organize this.
    for chart_id in chart_ids.values():
        frame = driver.find_element_by_css_selector(f'iframe[src*="//datawrapper.dwcdn.net/{chart_id}/"]')
        driver.switch_to.frame(frame)
        ...
```

And so does the function that gets the CSV data:

```py
def extract_csvs(chart_id: str, url: str) -> str:
    ...
    frame = driver.find_element_by_css_selector(f'iframe[src*="//datawrapper.dwcdn.net/{chart_id}/"]')

    driver.switch_to.frame(frame)
    # Grab the raw data out of the link's href attribute
    csv_data = driver.find_element_by_class_name('dw-data-link').get_attribute('href')
    ...
```

So it seemed like both need to switch frames, and both could share some code for that. For example:

```py
@contextmanager
def chart_frame(driver, chart_id):
    frame = driver.find_element_by_css_selector(f'iframe[src*="//datawrapper.dwcdn.net/{chart_id}/"]')
    driver.switch_to.frame(frame)
    try:
        yield frame
    finally:
        driver.switch_to.default_content()

def get_chart_data(driver, chart_id):
    with chart_frame(driver, chart_id):
        csv_data = driver.find_element_by_class_name('dw-data-link').get_attribute('href')
    # Do all the data parsing
    return result

def get_chart_meta(driver, chart_id):
    with chart_frame(driver, chart_id):
        notes = driver.find_elements_by_class_name('dw-chart-notes')
        return list({note.text for note in notes})
```

covid19_sfbayarea/data/marin_scraper.py
Mr0grog: No, sorry, I wasn’t clear. I was referring to the fact that `datetime.strptime()` returns an instance of `datetime.datetime` rather than `datetime.date`. And calling `.isoformat()` on a `datetime` gets you a string like `2020-05-12T00:00:00`, but we want a string like `2020-05-12`.

So if `row.split(',')[0]` was `'5/12/2020'` in this code, you’d get:

```py
>>> date_time_obj = datetime.strptime('5/12/2020', '%m/%d/%Y')
>>> date_time_obj.isoformat()
'2020-05-12T00:00:00'
```

You can fix that in a couple ways:

```py
# Convert to a datetime.date first:
date_time_obj.date().isoformat()

# Or provide an explicit format instead of isoformat:
date_time_obj.strftime('%Y-%m-%d')
```

kwonangela7: ah my bad - for whatever reason I thought we wanted a string like `2020-05-12T00:00:00`. I'll make sure to convert to a date time.date first, thanks Rob!

covid19_sfbayarea/data/marin_scraper.py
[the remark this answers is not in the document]
kwonangela7: OHHHH okay, that makes sense. I was misunderstanding what you were saying about get_chart_data in your first comment. Thanks for the tip Rob!

covid19_sfbayarea/data/marin_scraper.py
kwonangela7: Thanks for the tip Rob! I ended up taking your suggestion, and am trying to see if there's a way to do this that uses the csv module (right now I'm splitting by commas and like you've said, I should try to avoid that..)

kwonangela7:
Sorry, I just realized my comment about the testing data disappeared. The Marin County pos/neg testing breakdown is for testing done for Marin County residents only. There is data for the total number of tests done in the county, but there is no breakdown for that. I'll bring it up during the worksession today, but should we scrape both kinds of data?

Also, we should double check the other scrapers to see if other counties explicitly make that differentiation. Without that context as to whether it's total number of tests done in the county or done for county residents, the comparison of tests numbers won't make sense..

covid19_sfbayarea/data/marin.py
```diff
@@ -0,0 +1,277 @@
...
+from .utils import get_data_model
+
+def get_county() -> Dict:
+    """Main method for populating county data"""
+
+    url = 'https://coronavirus.marinhhs.org/surveillance'
+    model = get_data_model()
+
+    chart_ids = {"cases": "Eq6Es", "deaths": "Eq6Es", "inmates": "KCNZn", "age": "zSHDs", "gender": "FEciW", "race_eth": "aBeEd"}
+    # I removed "tests": '2Hgir' from chart_ids b/c it seems to have disappeared from the website?
```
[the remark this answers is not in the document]
elaguerta: You'll have to remove 'inmates', I just saw this on the site today:
"As of 8/24/2020 we will no longer report San Quentin numbers on this page. However, you can still track all of San Quentin data at the California Department of Corrections and Rehabilitation website which is the source of our data."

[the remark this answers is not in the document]
elaguerta: I think testing is back now: https://datawrapper.dwcdn.net/7sHQq/

covid19_sfbayarea/data/marin.py
```diff
@@ -0,0 +1,277 @@
...
+
+    #model["series"]["tests"] = get_test_series(chart_ids["tests"], url)
+    model["case_totals"]["age_group"], model["death_totals"]["age_group"] = get_breakdown_age(chart_ids["age"], url)
+    model["case_totals"]["gender"], model["death_totals"]["gender"] = get_breakdown_gender(chart_ids["gender"], url)
+    model["case_totals"]["race_eth"], model["death_totals"]["race_eth"] = get_breakdown_race_eth(chart_ids["race_eth"], url)
+    return model
+
+@contextmanager
+def chart_frame(driver, chart_id: str): # type: ignore 
+    # is this bad practice? I didn't know what type to specify here for the frame.
```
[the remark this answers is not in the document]
elaguerta: I'd probably use `Any` just to pass the linter but I am the kind of person who just wants to pass the linter

covid19_sfbayarea/data/marin.py
```diff
@@ -0,0 +1,277 @@
...
+    genders = ['male', 'female']
+    c_gender: dict = dict()
+    d_gender: dict = dict()
+    
+    for row in csv_reader:
+        # Extracting the gender and the raw count (the 3rd and 5th columns, respectively) for both cases and deaths.
+        # Each new row has data for a different gender.
+        gender = row["Gender"].lower()
+        if gender not in genders:
+            return ValueError("The genders have changed.") # type: ignore
```
elaguerta, line 216:
```suggestion
          raise ValueError("The genders have changed.") # type: ignore 
```

data_models/README.md
```diff
@@ -137,7 +138,18 @@ The fields will be used for normalizing the county case and death tabulations, a
     }
 ```
 
-5. __Hospitalization Data__
+6. __Inmate Data__
+
```
elaguerta, line 142: You could update this part to say that data collection is pending resolution of #108.

data_models/README.md
```diff
@@ -205,6 +217,8 @@ Scraper authors, please keep an eye out for amendments to the data model.
 # Race and Ethnicity
 We need to collapse counties that report race and ethnicity into one race/ethnicity dimension. This section will be updated pending information about San Francisco County's methods for reporting race and ethnicity.
 
+The category "Multi_or_Other" was included because Marin rolls up the numbers from "Multi" and "Other" into one.
+
```
elaguerta, line 221: Maybe add: "Please note that this category is not relevant for counties that report 'Mulitple Race' and 'Other Race' separately."
This is yet another sign that maybe we should just take the race category names exactly as the county reports them, like we're now doing with gender and age.

ldtcooper: Honestly, this makes a lot of sense to do. It would come with the bonus of making all the scrapers more resilient against changes to the reported races like what happened with Sonoma

kwonangela7: I'll add that Elaine - thank you both!

kwonangela7:
Thank you @elaguerta and @ldtcooper for your comments! @elaguerta unfortunately the original negative/positive breakdown for tests is gone, but I'll go ahead and populate the date/positive/cumul_post since that's what I can scrape from the graph you linked. Thanks for the update!

kwonangela7:
@elaguerta I think I covered all of your requested changes, and made sure that I used 4 space indents like Rob requested above. I also added some discussion points in our data fetch channel, and edited the metadata

Let me know if you think it's ready to merge!

elaguerta:
Looks good to me. Thanks for your contributions!

elaguerta (approved):

PR #86, opened 2020-06-17
Clean up language and date issues in Contra Costa News

Mr0grog:
I noticed a couple issues in the [Contra Costa news scraper's updates](https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/186/files) today:

1. Some titles had “| Spanish” at the end because some news items have alternate language versions, and we weren’t pulling those out from the title. (The links were still fine.)

2. Some dates on the page have typos that caused them to get parsed poorly. 😞 

This fixes both of those.

### DIFF
--- a/covid19_sfbayarea/news/utils.py
+++ b/covid19_sfbayarea/news/utils.py
@@ -11,7 +11,9 @@ from .feed import NewsItem
 
 HEADING_PATTERN = re.compile(r'h\d')
 ISO_DATETIME_PATTERN = re.compile(r'^\d{4}-\d\d-\d\d(T|\s)\d\d:\d\d:\d\d(\.\d+)?(Z|\d{4}|\d\d:\d\d)$')
+US_SHORT_DATE_PATTERN = re.compile(r'^\s*\d+/\d+/\d+\s*$')
 PACIFIC_TIME = dateutil.tz.gettz('America/Los_Angeles')
+CURRENT_YEAR = datetime.utcnow().year
 
 # Sometimes we can't find a news feed that is specific to COVID-19, so the news
 # items we scrape need to be filtered. These key terms are used to test whether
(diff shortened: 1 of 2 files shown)


covid19_sfbayarea/news/utils.py
```diff
@@ -114,6 +114,10 @@ def parse_datetime(date_string: str, timezone: Optional[tzinfo] = PACIFIC_TIME)
     the `timezone` argument to set the timezone to use if none was specified in
     the parsed string.
     """
+    # Handle dumb typos that might be in the dates on the page :(
+    if date_string.endswith('/202'):
+        date_string += '0'
+
```
elaguerta, line 120: While we're here, you could check that date_string is in the form `"mm/dd/yyyy"`, or check against a set of date patterns the news sources use, and raise an error if it's not. Due to the typos in Contra Costa's source, I think we should leave `ARTICLE_TITLE_PATTERN` as it is and do a stricter check here.
```suggestion
    # Handle dumb typos that might be in the dates on the page :(
    DATE_PATTERN = re.compile(r'^\d{2}/\d{2}/\d{4}$')     # The date in mm/dd/yyyy format. 
    if not DATE_PATTERN.match(date_string):
        # start a list of if's to handle every typo we might find, starting with the '/202' typo
        if date_string.endswith('/202'):
            date_string += '0'
        else:
            raise FormatError("Date string does not match pattern 'dd/mm/yyyy' ")
```

Mr0grog: Hmmmm, if we did a check like this, it would have to move out of this function and directly into the Contra Costa scraper (this function handles a wide variety of formats; I thought this check was safe because I can’t think of any format that we would expect to end with “/202” or “/2020” besides mm/dd/yyyy).

An alternative might be to keep the check here, but not raise an error:

```suggestion
    DATE_PATTERN = re.compile(r'^\d+/\d+/\d+$')
    if DATE_PATTERN.match(date_string):
        if date_string.endswith('/202'):
            date_string += '0'
```

Or to check and fix the year *after* parsing (would work for this case, not sure what other odd future typos would be fixable this way):

```py
def parse_datetime(date_string: str, timezone: Optional[tzinfo] = PACIFIC_TIME) -> datetime:
    date = dateutil.parser.parse(date_string)

    # Handle a possible typo in the date string.
    if date.year == 202:
        date = date.replace(year=2020)

    # Gut-check whether this date seems reasonable
    if abs((datetime.datetime.now(date.tzinfo) - date).days) > 5 * 365:
        raise ValueError(f'Unknown date format: "{date_string}"')

    if date.tzinfo is None:
        date = date.replace(tzinfo=timezone)

    return date
```

Mr0grog: Updated this to a mix of the above approaches. I’m going to go ahead and merge since you already approved here, @elaguerta, but please comment if you think it should be revised further in a new PR.

PR #89, opened 2020-06-26
Ensure reliable order for items in news feeds

### LINKED ISSUE #88: Maintain stable sort order for news items
While checking the [latest news scraper output](https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/196/files), I noticed that sometimes news items are showing up in different orders from one scrape to the next. This is generally a result of sites that only surface the publish *date* and not the publish *time*, and that have more than one news item on a given day. We sort based on publish time: https://github.com/sfbrigade/data-covid19-sfbayarea/blob/47ced984c3379fa4b5908e3de1e9778714d5e7c4/covid19_sfbayarea/news/feed.py#L92-L94

The most sensible thing to do here is probably to fall back on other sorting keys when the time is the same, such as `title`, `url`, or `id` (although `id` is often the same as `url` for us, since underlying primary keys, etc. are often not surfaced).

Mr0grog:
In situations where multiple feed items have the same publish time (usually because the county only specifies the date and not the time), we sometimes wind up with feed items getting a slightly different order between scraper runs. We now use the ID of the news items as an additional sort criteria to make sure they stay in the same order.

This change also adds some actual tests!

Fixes #88.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,40 @@ To run the scraper, you can use the run script by typing `sh run_scraper.sh` int
 
 ## Running the API
 The best way to run the API right now is to run the command `FLASK_APP="app.py" FLASK_ENV=development flask run;`. Note that this is not the best way to run the scraper at this time.
+
+## Development
+
+We use CircleCI to lint the code and run tests in this repository, but you can (and should!) also run tests locally.
+
+The commands described below should all be run from within the virtual environment you’ve created for this project. If you used `install.sh` to get set up, you’ll need to activate your virtual environment before running them with the command:
+
+```sh
+$ source env/bin/activate
+```
+
+If you manage your environments differently (e.g. with Conda or Pyenv-Virtualenv), use whatever method you normally do to set up your environment.
+
+### Tests
+
+You can run tests using `pytest` like so:
+
+```sh
+# In the root directory of the project:
+$ python -m pytest -v .
+```
+
+### Linting and Code Conventions
+
+We use Pyflakes for linting. Many editors have support for running it while you type (either built-in or via a plugin), but you can also run it directly from the command line:
+
+```sh
+# In the root directory of the project:
+$ pyflakes .
+```
+
+We also use type annotations throughout the project. To check their validity with Mypy, run:
+
+```sh
+# In the root directory of the project:
+$ mypy .
+```
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,15 +8,15 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - data-covid-19-sfbayarea-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
-            - data-covid-19-sfbayarea-{{ arch }}-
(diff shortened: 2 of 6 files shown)


Mr0grog:
Wow, that got a little messy with CircleCI! We had some issues where `install.sh` was being run incorrectly in a way that *worked*, but now how it was supposed to. That happened to be fine with the way our linting scripts were written, but broke with how the pytest tests were written. It’s fixed now, though.

In other news, I learned that:

```sh
$ sh install.sh
```

Can’t actually create a virtual environment on the version of Linux we’re using in CircleCI! So what was happening was that we installed all our dependencies into the global environment, not the virtual environment. Instead, either of the following methods work:

```sh
# Use bash instead of sh:
$ bash install.sh
# Or just run it directly, since we made it executable a while back:
$ ./install.sh
```

**If you originally set up your local project with `sh install.sh`, you might want to make sure it did the the right thing on your own machine.**

README.md
```diff
@@ -10,3 +10,40 @@ To run the scraper, you can use the run script by typing `sh run_scraper.sh` int
...
+
+If you manage your environments differently (e.g. with Conda or Pyenv-Virtualenv), use whatever method you normally do to set up your environment.
+
+### Tests
+
+You can run tests using `pytest` like so:
+
+```sh
+# In the root directory of the project:
+$ python -m pytest -v .
```
elaguerta, line 32: I got a couple warnings from running pytest on root. Are these ok?
```
env/lib/python3.7/site-packages/html5lib/_trie/_base.py:3
  /Users/elainelaguerta/OpenSourceProjects/data-covid19-sfbayarea/env/lib/python3.7/site-packages/html5lib/_trie/_base.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Mapping

covid19_sfbayarea/news/utils.py:42
  /Users/elainelaguerta/OpenSourceProjects/data-covid19-sfbayarea/covid19_sfbayarea/news/utils.py:42: DeprecationWarning: invalid escape sequence \?
    b'<?xml\\s[^>]*encoding=[\'"]([^\'"]+)[\'"].*\?>',
```

Mr0grog: Well, they’re both warnings, so technically fine. 😉 

But anyway, the first one is coming from `env/lib/python3.7/site-packages/html5lib/_trie/_base.py`, i.e. it’s coming from a library we depend on, not from us. (That said, [it looks like it’s been fixed in v1.1](https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#11), which just came out last week, so we should consider upgrading. I think that’s out of scope for this PR, though.)

The second one is new in Python 3.8, but is a good warning I should probably fix.

Mr0grog: Actually, that’s not new in 3.8, but I’m not getting it locally for some reason. Odddddddd.

Mr0grog: Ah, Pytest was being helpful by not repeating warnings it had already shown me on files I hadn’t edited. Ha. Anyway, fixed.

PR #90, opened 2020-06-26
Add shebang line to executable bash scripts

Mr0grog:
Back in #54, @rickpr made the `.sh` scripts and top-level `.py` scripts in this project executable. However, the bash scripts don't have a shebang line describing what interpreter should be used. Thankfully, most systems default to bash in this case, so everything usually works OK -- but that's no guarantee. This makes things a little safer by adding a shebang line specifying that bash should be used.

(Inspired by the small mess I ran into in #89.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 python3 -m venv env;
 source env/bin/activate;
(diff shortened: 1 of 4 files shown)


Mr0grog:
Whoa, I did not know that, either!

PR #91, opened 2020-06-26
Limit news scrapers to a given timeframe

### LINKED ISSUE #43: Limit news scraper to time range
Over on the front-end side, @kengoy noted that we might want to [limit the time range](https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/79#issuecomment-621010886) included in the news feeds we produce:

> Another consideration is we may want to filter only for the new feeds like for the past 1 or 2 weeks instead of showing all the past feeds.

The current SF scraper just takes everything on the first page of news. Instead, [`NewsScraper`](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/ac696a1444649a1187dbda8e2dea5635e2d47953/covid19_sfbayarea/news/base.py#L7-L61) should take a time range or a “from” date to limit what news items get returned.

---

*Update 2020-06-11: Fixed `NewsScraper` link to point at the class’s new location since the code has been refactored a few times.*

Mr0grog:
Use the new `--from` CLI argument to set a timeframe for news. It can be a date after which news items should be included OR a number of days before the current time:

```sh
# Scrape all news after June 5, 2020
$ python scraper_news.py --from 2020-06-05

# Scrape all news after one week ago
$ python scraper_news.py --from 7
```

The default is "31" -- i.e. the past month.

Fixes #43.

### DIFF
--- a/covid19_sfbayarea/news/base.py
+++ b/covid19_sfbayarea/news/base.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import requests
 from typing import Dict, List
 from .feed import NewsFeed, NewsItem
@@ -22,6 +23,10 @@ class NewsScraper:
     FEED_INFO: Dict = {}
     URL = ''
 
+    def __init__(self, from_date: datetime = None, to_date: datetime = None) -> None:
+        self.from_date = from_date
+        self.to_date = to_date or datetime.now().astimezone()
+
     def create_feed(self) -> NewsFeed:
         return NewsFeed(**self.FEED_INFO)
 
@@ -44,7 +49,9 @@ class NewsScraper:
         feed = self.create_feed()
         html = self.load_html(self.URL)
         news = self.parse_page(html, self.URL)
-        feed.append(*news)
+        feed.append(*(item
+                      for item in news
+                      if self._in_time_range(item)))
         return feed
 
     def load_html(self, url: str) -> str:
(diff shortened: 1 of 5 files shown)


covid19_sfbayarea/news/base.py
```diff
@@ -55,7 +62,12 @@ def load_html(self, url: str) -> str:
     def parse_page(self, html: str, url: str) -> List[NewsItem]:
         raise NotImplementedError()
 
+    def _in_time_range(self, candidate: NewsItem) -> bool:
+        time = candidate.date_published
+        return time < self.to_date and (not self.from_date or
```
elaguerta, line 67: Should the first comparison be
`time <= self.to_date` ?

Mr0grog: Oh, this was intentional, but I’m happy to change it.

Mr0grog:
@elaguerta updated the time range to be inclusive of the end time. Ready for a re-review.

covid19_sfbayarea/news/base.py
Mr0grog: (Worth noting that everything here is actually a `datetime`, so if `to_date` was specified, we’re not saying “before today,” we’re saying “before right now.”)

elaguerta: I didn't catch that, sorry! I thought it would return `before today`. Well, we really will have `only the latest news`. Thanks for changing it.

Mr0grog:
@ldtcooper how’s that?

PR #94, opened 2020-07-02
Default to *all* counties for news scraper CLI

### LINKED ISSUE #92: News scraper should default to *all* counties
If you don’t specify what counties you want, the news scraper currently defaults to only scraping San Francisco:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/eefce7a08ba7ed0c9ac9655b69a0f1588ec9d293/scraper_news.py#L21-L24

This is an artifact of how the script was originally set up for use in the frontend repo, and we had to do it for backwards compatibility. However, the frontend now explicitly specifies all counties, so it should be safe to update the default.

Mr0grog:
If you didn't specify any county names to scrape, the news scraper used to only output news for San Francisco. It now outputs news for all counties instead, which is a more sensible default and matches up with the data side. (Code that depended on the old default has now all been fixed, so this change is finally safe.)

While we’re at it, this also updates the default format from simple JSON to JSON feed.

Fixes #92.

### DIFF
--- a/scraper_news.py
+++ b/scraper_news.py
@@ -3,25 +3,23 @@ import click
 import json
 from covid19_sfbayarea import news
 from pathlib import Path
-from typing import Tuple
+from typing import cast, Tuple
 
 
-COUNTY_NAMES = tuple(news.scrapers.keys())
+COUNTY_NAMES = cast(Tuple[str], tuple(news.scrapers.keys()))
 
 
 @click.command(help='Create a news feed for one or more counties. Supported '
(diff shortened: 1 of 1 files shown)


Mr0grog:
> it wrote separate `.json` files for each county. That's different from the data side, where we're writing one long `.json`.

Yep, that’s intentional. These follow the JSON Feed spec so they are subscribable in news readers, but that also means we can’t have separate sections in one file for each county.

PR #104, opened 2020-07-10
Sf broke on gender its a feature not a bug

elaguerta:


elaguerta:
Strange error from circleci. I think it has something to do with this line in `.install.sh`:
`webdrivermanager firefox --linkpath "$(pwd)/env/bin";`

image

covid19_sfbayarea/data/san_francisco.py
```diff
@@ -152,12 +153,16 @@ def get_age_table(session : SocrataApi, resource_ids: Dict[str, str]) -> List[Di
     AGE_KEYS = {"18_and_under": "under 18", "18_to_30": "18-30", "31_to_40": "31-40", "41_to_50": "41-50",
                 "51_to_60": "51-60", "61_to_70": "61-70", "71_to_80": "71-80", "81_and_older": "81+",}
 
+    # find the latest date of data collection
+    params = {'$select': 'max(specimen_collection_date) as date'}
+    latest_date = session.resource(resource_id, params=params)[0]
```
Mr0grog, line 158: I think this all might read a bit more clearly if `latest_date` was the actual date string instead of a dict:

```suggestion
    latest_date = session.resource(resource_id, params=params)[0]['date']
```

covid19_sfbayarea/data/san_francisco.py
```diff
@@ -168,67 +173,65 @@ def get_age_table(session : SocrataApi, resource_ids: Dict[str, str]) -> List[Di
...
-
-    # re-key
-    gender_table = {GENDER_KEYS[entry["gender"]]: int(
-        entry["cumul_cases"]) for entry in data}
-
-    return gender_table
+    GENDER_KEYS = {"Female": "female", "Male": "male", "Unknown": "unknown", "Trans Female": "female", "Trans Male": "male"}
+    # find the latest date of data collection
+    params = {'$select': 'max(specimen_collection_date) as date'}
+    latest_date = session.resource(resource_id, params=params)[0]
```
Mr0grog, line 182: Same here as above about making this the date string instead of a dict:

```suggestion
    latest_date = session.resource(resource_id, params=params)[0]['date']
```

covid19_sfbayarea/data/san_francisco.py
```diff
@@ -168,67 +173,65 @@ def get_age_table(session : SocrataApi, resource_ids: Dict[str, str]) -> List[Di
...
+    params = {'$select': 'gender, cumulative_confirmed_cases as cases', '$where':f'specimen_collection_date="{latest_date["date"]}"'}
     data = session.resource(resource_id, params=params)
+
     # re-key
-    transmission_data = { TRANSMISSION_KEYS[ entry["transmission_category"] ]: int(entry["sum_case_count"]) for entry in data}
-    return transmission_data
+    table : Dict[str, int]= dict()
+    for entry in data:
+        rekey = GENDER_KEYS[entry['gender']]
+        table[rekey] = table.get(rekey,0) + int(entry["cases"])
```
Mr0grog, line 192: This threw me for a minute, making me think you were working around some case where the source data had repeated genders, but then I realized it’s because we are combining some of the genders from the source data. It would probably be useful to have a comment here explaining that’s why you’re doing this otherwise-slightly-obtuse-looking operation.

covid19_sfbayarea/data/san_francisco.py
```diff
@@ -168,67 +173,65 @@ def get_age_table(session : SocrataApi, resource_ids: Dict[str, str]) -> List[Di
...
     """
-    resource_id = resource_ids["race_eth"]
+    resource_id = resource_ids['race_eth']
     # Dict of source_label:target_label for re-keying.
-
     RACE_ETH_KEYS = {'Hispanic or Latino/a, all races': 'Latinx_or_Hispanic', 'Asian': 'Asian', 'Black or African American': 'African_Amer', 'White': 'White',
                     'Native Hawaiian or Other Pacific Islander': 'Pacific_Islander', 'Native American': 'Native_Amer', 'Multi-racial': 'Multiple_Race', 'Other': 'Other', 'Unknown': 'Unknown'}
+    # find the latest date of data collection
+    params = {'$select': 'max(specimen_collection_date) as date'}
+    latest_date = session.resource(resource_id, params=params)[0]
```
Mr0grog, line 206: Same as above w/r/t extracting the date.

```suggestion
    latest_date = session.resource(resource_id, params=params)[0]['date']
```

Actually, given how much you’re repeating this, is it worth abstracting into a function you can call from all the places you do it?

```py
def get_latest_specimen_date(resource_id: str) -> str:
    data = session.resource(resource_id, params={
        '$select': 'max(specimen_collection_date)'
    })
    return data[0]['date']
```

PR #105, opened 2020-07-11
faq-json parser

zappascout:
Automates the parsing of text from UX team's FAQ Contents spreadsheet to JSON, ready for an Action to complete the automation process. Rob's review notes taken in. A future iteration may make use of DictReader. Type annotation added :-)

### DIFF
--- a/jsonic3_commit.py
+++ b/jsonic3_commit.py
@@ -1,0 +1,80 @@
+"""
+Grabs data from the 'FAQ Content' CSV and turns it into nice JSON: a main faq object containing an array of Section objects, each Section Object in turn holding an array of Question objects consisting of question, answer and related link/s, as so:
+    [
+        {
+          'Q': 'What are the symptoms of COVID-19?',
+          'A': 'Symptoms of COVID-19 include coughing and shortness of breath. Additionally, a person showing two or more of the following symptoms may have the virus: fever, repeated shaking with chills, muscle pain, headache, sore throat, new loss of taste or smell.',
+          'link': 'https://www.cdc.gov/coronavirus/2019-ncov/symptoms-testing/symptoms.html'
+        },
+    .
+    .
+    .
+    ]
+    etc
+In addition to an array of Question objects, each Section incorporates its title and a 'last updated' date value. For the Python intermediary stage (between CSV stream  and JSON) we will create a hierarchy of nested dictionaries and lists equivalent to the JSON objects and arrays.
+"""
+
+import requests
+import json
+import csv
+import datetime
+from typing import Iterable, Dict, List, Any
+
+
+# function to grab data from a google sheet + return as reader object
+def google_sheet_csv_data(sheet: str, gid: str) -> Iterable[List[str]]:
+    url = f'https://docs.google.com/spreadsheets/d/{sheet}/export'
+    response = requests.get(url, params={
+        'format': 'csv',
+        'id': sheet,
+        'gid': gid
+    })
+    response.raise_for_status()
+    return csv.reader(response.iter_lines(decode_unicode=True))
+
+
+# Pass faq Content sheet id and gid to function
+_reader = google_sheet_csv_data('1_wBXS62S5oBQrwetGc8_-dFvDjEmNqzqHwUeP-DzkYs',
+ '1318925039')
+
+# Date, to be updated each time the program runs, in UTC
+date = datetime.datetime.utcnow().strftime('%Y-%m-%d')
+
+# Create the main FAQ dictionary
+# Create a list to contain our Section objects
+# Place the Sections list inside the main FAQ dictionary
+faq_dict = {}
+sections_list: List[Dict] = []
+faq_dict['faqItems'] = sections_list
+
+# Ensure only first two columns in each row are used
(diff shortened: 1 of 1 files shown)


jsonic3_commit.py
```diff
@@ -0,0 +1,80 @@
...
+        'id': sheet,
+        'gid': gid
+    })
+    response.raise_for_status()
+    return csv.reader(response.iter_lines(decode_unicode=True))
+
+
+# Pass faq Content sheet id and gid to function
+_reader = google_sheet_csv_data('1_wBXS62S5oBQrwetGc8_-dFvDjEmNqzqHwUeP-DzkYs',
+ '1318925039')
```
Mr0grog, line 38: This is an absurdly minor formatting nit, but when spreading arguments, list items, dictionary entries, etc. over multiple lines I try to stick to one of two conventions for better readability that are fairly common in Python:

1. Align all the items:

    ```suggestion
    _reader = google_sheet_csv_data('1_wBXS62S5oBQrwetGc8_-dFvDjEmNqzqHwUeP-DzkYs',
                                    '1318925039')
    ```

2. Give each item its own line and indent one level:

    ```suggestion
    _reader = google_sheet_csv_data(
        '1_wBXS62S5oBQrwetGc8_-dFvDjEmNqzqHwUeP-DzkYs',
        '1318925039'
    )
    ```

jsonic3_commit.py
```diff
@@ -0,0 +1,80 @@
...
+
+# Date, to be updated each time the program runs, in UTC
+date = datetime.datetime.utcnow().strftime('%Y-%m-%d')
```
Mr0grog, line 41: I don’t remember if I remarked on this before, but would it be good to include the time here so it’s clear when updates were if we ultimately run them more than once a day?

If the front-end will choke if time is included, then obviously disregard that entirely. 😅

jsonic3_commit.py
```diff
@@ -0,0 +1,80 @@
...
+    elif rowtype == '':   # Some safeguards
+        pass
+    elif rowval and not rowtype:
+        raise ValueError(f'row {index} has a value but no description!')
+    else:
+        raise ValueError(f'Unknown row header: "{rowtype}"')
+
+# Create formatted json file
+with open('faq.json', 'w') as f:   # Create JSON output file
+    json.dump(faq_dict, f, indent=2)   # Put FAQ Content in output file as JSON
```
Mr0grog, line 80: I generally find scripts like this to be much more useful and flexible if they print their result to STDOUT (i.e. `print(result)` or `sys.stdout.write(result)`) rather than writing to a predetermined file. That way you can put the results wherever you want when running it:

```sh
$ python jsonic3_commit.py > path/where/i/want/faq.json
```

When you’ve got multiple outputs or more complex output, you can provide a `-o` or `--output` option for specifying where to write the output, like we do with `scraper_data.py` or `scraper_news.py`:
https://github.com/sfbrigade/data-covid19-sfbayarea/blob/06342ce29cf95b68c115cf0660ec904489e7f5b7/scraper_news.py#L18-L19
https://github.com/sfbrigade/data-covid19-sfbayarea/blob/06342ce29cf95b68c115cf0660ec904489e7f5b7/scraper_news.py#L38-L44

But that’s not a hard requirement. I just think it tends to work a lot better.

zappascout:
Hey Rob

Welcome back!

Thanks hugely for the review notes on the faq parser - all noted and will look forward to getting stuck into them after tonight’s meeting.

Thanks especially for the ‘formatting nits:' I think these things are priceless, as the joy of Python for me is its aesthetic. Funnily enough, the second of your suggestions for dealing with multiline arguments is very familiar to me from my first language, SuperCollider, which is based on C and Smalltalk and feels like a messy cross between Ruby and Javascript - it’s always so nice to look at Python after I've spent time with SC!

Anyway, thanks again Rob, and hope you had a restful break.

asx

@wiresmith
http://andrewsmithauthor.com <http://andrewsmithauthor.com/>

jsonic3_commit.py
zappascout: Actually, not too sure about this one, as the faqs are written by members of the team rather than scraped from somewhere else, so the likelihood of ever needing to run this twice a day automatically is pretty remote. If some piece of information changed and needed to be updated urgently, we'd probably just run the script manually. 
Also think Kengo will have kittens if we suggest adding time! :-)

jsonic3_commit.py
zappascout: Yeah, this is great - thanks v much for the steer...enjoyed reading all the docs and learning this and will def action.

zappascout:
Lots learnt: will action and resubmit under new improved name!

Mr0grog:
FWIW: in the future, you can always just `git mv jsonci2_commit.py new/file/name.py` and commit that, rather than closing and starting a new PR.

Mr0grog:
(And if you’re worried about keeping a clean history, you can always select “squash and merge” [which merges the PR as a single commit] instead of just selecting “merge.”)

zappascout:
brilliant - thanks for that, will do that way next time :-)



@wiresmith
http://andrewsmithauthor.com <http://andrewsmithauthor.com/>

PR #111, opened 2020-08-06
Add Dockerfile that works for development use

Mr0grog:
Since it can be tough to get all the pieces installed and working right when using WSL on Windows, this dockerfile provides a way to develop and run the app in a container with all the necessary dependencies. You can build an image from it with:

    $ docker build -t data-covid19-sfbayarea .

And then run it via:

    # Get a bash prompt inside the container:
    $ docker run -it data-covid19-sfbayarea

    # Or instead just run the scraper in the
    # container and quit the container when the
    # scraper finishes:
    $ docker run -it data-covid19-sfbayarea python scraper_data.py --more args --here

@ldtcooper does this work for you? It might be better to test with the news scraper, since it seems like some of the data scrapers are broken at the moment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+# This Docker image setup is designed primarily for development use. You *can*
+# use it to run the scraper tools in a "production" capacity, but it's a lot
+# heavier weight, with extra tools & dependencies, than you'll need.
+FROM python:3.8-slim-buster
+
+# Debian does not support a stable package for Firefox. Instead, it has
+# "Firefox-ESR (Extended Support Release)", which is not very up-to-date, and
+# does not seem to work too well with geckodriver. (It might be possible, but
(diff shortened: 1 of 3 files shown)


ldtcooper:
It seems to be working in that I'm no longer getting Selenium errors, and am instead getting Python errors whenever I try to run either scraper

Mr0grog:
Were the news scraper errors #113? If so, seems like this is working great.

ldtcooper:
This works for me!

Mr0grog:
@ldtcooper how does this look? Added a quick helper script plus [docs](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/dockerize-for-easier-windows-development/README.md#using-docker).

ldtcooper (approved):
This looks great! I'm looking forward to using it

PR #113, opened 2020-08-13
Update Alameda news scraper to handle new HTML

Mr0grog:
Alameda county redesigned their COVID-19 site, moving news to a different URL and changing the HTML layout of the page. This updates the news scraper to handle the new format, which is a little more complicated to parse. (On the upside, it no longer requires Selenium.)

### DIFF
--- a/covid19_sfbayarea/news/alameda.py
+++ b/covid19_sfbayarea/news/alameda.py
@@ -1,11 +1,10 @@
-from bs4 import BeautifulSoup, element  # type: ignore
-from typing import List
+from bs4 import BeautifulSoup, element, NavigableString  # type: ignore
+from typing import List, Optional
 from urllib.parse import urljoin
-from ..webdriver import get_firefox
 from .base import NewsScraper
 from .errors import FormatError
 from .feed import NewsItem
-from .utils import first_text_in_element, get_base_url, parse_datetime
+from .utils import get_base_url, parse_datetime
 
 
 class AlamedaNews(NewsScraper):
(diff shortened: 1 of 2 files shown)


Mr0grog:
Should be fixed.

covid19_sfbayarea/news/alameda.py
```diff
@@ -56,30 +56,98 @@ def load_html(self, url: str) -> str:
...
+        #     <strong>July 17, 2020</strong>
+        #     The title of the news article
+        #     <a href="/path/to/english/article.html">English</a> |
+        #     <a href="/path/to/spanish/article.html">Spanish</a> |
+        #     <br>
+        #
+        url = None
+        title = ''
+        for sibling in node.next_siblings:
+            # <br> tags signal the end of the news item.
```
elaguerta, line 131: Alas, this no longer seems to be true.
image
I think there are 2 `<br>` tags separating each news item (for now) so you could check for two in a row. Or you could check for the next `<strong>` tag. I was going to write a suggestion to break after seeing a `strong` tag but for some reason that resulted in just parsing the August news items. I confirmed that all the `strong` tags pass line 110.

Mr0grog: JFC. Looking at the second news item, the link’s `title` attribute implies that the `<br>` tag inside the news item is delimiting the title from a lede or subhead. So I think I’m going to take everything up to the `<br>` as the title and everything after as summary.

More fun is that the second news item has no English version, even thought the title is in English. I think I’ll treat that like non-English entries in other news feeds and not surface it. That’s definitely not ideal, but I feel like it fits best with how we’ve been treating things.

Also! Down in April, most of the entries don’t start with a `<strong>` tag at all! Since, practically speaking, we are never going to show content from that far back, I don’t think I’m going to worry about it. But still, ugh.

Mr0grog:
Ugh, OK, I rewrote the scraper as a finite state machine, which felt like a much more sane way to handle all the crazy conditions possible here. How's this look to you, @elaguerta?

I think I am also going to update it to include news items that have no English version, but the way I am thinking of doing that would add a whole bunch more changes, so I’m going to put that in a separate PR.

elaguerta (approved):
Looks good! I dig the finite-state-machine approach!

PR #115, opened 2020-08-19
Sonoma: Fix Racial Groups

ldtcooper:
It looks like Sonoma has finally had enough mixed race and African-American cases to warrant taking them out of the Other category. They've also moved Other out of a footnote and into the table. Both of these things broke the scraper.


ldtcooper:
Git is being a huge pain. I'm going to make this PR again on a different branch

PR #118, opened 2020-09-03
Let data scrapers fail independently

rickpr:
Send a webhook to Slack if any failures occur.

### DIFF
--- a/scraper_data.py
+++ b/scraper_data.py
@@ -16,7 +16,7 @@ COUNTY_NAMES : Tuple[str,...]= tuple(data_scrapers.scrapers.keys())
                 type=click.Choice(COUNTY_NAMES, case_sensitive=False))
 @click.option('--output', metavar='PATH',
               help='write output file to this directory')
-def main(counties: Tuple[str,...], output: str) -> int:
+def main(counties: Tuple[str,...], output: str) -> None:
     out = dict()
     failed_counties = False
     if len(counties) == 0:
@@ -39,10 +39,8 @@ def main(counties: Tuple[str,...], output: str) -> int:
     else:
         print(json.dumps(out,indent=2))
 
-    if not out: # all counties failed
-        exit(70)
-    elif failed_counties: # some counties failed
-        exit(1)
+    if not out: exit(70) # all counties failed
+    if failed_counties: exit(1) # some counties failed
 
 if __name__ == '__main__':
     main()


scraper_data.py
```diff
@@ -21,8 +22,14 @@ def main(counties: Tuple[str,...], output:str) -> None:
...
     # Run each scraper's get_county() method. Assign the output to out[county]
+    failed_counties = []
     for county in counties:
-        out[county] = data_scrapers.scrapers[county].get_county()
+        try:
+            out[county] = data_scrapers.scrapers[county].get_county()
+        except:
+            failed_counties.append(county)
+
+    if failed_counties: send_webhook(failed_counties)
```
Mr0grog, line 32: Quick ’n dirty thoughts as you are refining this tonight:

1. We should still log the error to stderr, either here or above in the `except` clause.

2. We should exit with a non-zero code to signify that things failed (e.g. `sys.exit(1)`). I think it might also be nice to have different exit codes for “all counties failed” (so we got nothing) and “some counties failed” (so we got something worth using, but there were still issues).

scraper_data.py
```diff
@@ -33,5 +39,13 @@ def main(counties: Tuple[str,...], output:str) -> None:
     else:
         print(json.dumps(out,indent=2))
 
+    if not out: # all counties failed
+        return 126
```
Mr0grog, line 43: In some shells, 126 and 127 have special meanings, so this one might not be a great choice. Other than that, things are pretty free-form and the shell won’t do anything special with the code, so 1-125 is usually safe.

(If you really want to finesse it, many \*nix systems define some common codes in `/usr/include/sysexits.h`, which you can check out. There’s not much out there that really treats programs using those specially, though. It’s more of a loose standard.)

rickpr: Went with `70` as it seems appropriate for a complete failure: https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD%204.3-RELEASE&format=html

scraper_data.py
```diff
@@ -33,5 +39,13 @@ def main(counties: Tuple[str,...], output:str) -> None:
...
+        return 70
+    elif failed_counties: # some counties failed
+        return 1
+    else: # success!
+        return 0
+
 if __name__ == '__main__':
-    main()
+    exit_code = main()
+    exit(exit_code)
```
Mr0grog, line 51: Ah, this doesn’t work for me, and I’m pretty sure it’s because Click commands usually don’t actually return their return value (see https://click.palletsprojects.com/en/7.x/commands/#command-return-values).

It’s probably better to just call `sys.exit(code)` up on lines 43 and 45.

rickpr: Makes sense - calling `sys.exit` on lines 43 and 45 instead to avoid getting caught up by click!

Mr0grog (approved):
👍 👍 👍

PR #120, opened 2020-09-05
Load more news from Solano via fancy XHR endpoint

Mr0grog:
The main news page for Solano County presents only a limited number of items, and doesn't give us enough to actually cover the full time range we're looking for in a lot of cases. Instead of loading and parsing the main news page, we now load the same endpoint that is used to display the county's full news archives.

NOTE: this required some changes to response body decoding since the endpoint does not specify the encoding it uses. I had to add a way to set a hard-coded fallback encoding when one cannot be inferred from the HTTP response.

### DIFF
--- a/covid19_sfbayarea/news/base.py
+++ b/covid19_sfbayarea/news/base.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 import requests
-from typing import Dict, List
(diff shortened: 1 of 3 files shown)


rickpr (approved):
Tested locally - output looks good and the new page makes sense!

PR #122, opened 2020-09-10
Let news scrapers fail independently

Mr0grog:
This ports the changes to the data scrapers from #118 over to the news scrapers, adds slightly nicer formatting, and includes a stack trace. The main thing is that an exception in any scraper does not stop the others from completing. Output-wise, if the Alameda scraper failed, you'd see something like:

Screen Shot 2020-09-09 at 8 13 35 PM

I added the stack trace in because it seemed like that would be pretty useful information most of the time, but we could also:

- Only show it for errors that aren’t `FutureWarning` or `FormatError` (on the assumption that those errors were expected and are easy to track down, while others are a surprise and the stack would be doubly helpful).
- Add a CLI option to show it, e.g. `--verbose` or `--trace-errors`.
- Add an environment variable to show it, e.g. `TRACE_ERRORS=true`.
- Just strip it back out entirely.

### DIFF
--- a/covid19_sfbayarea/utils.py
+++ b/covid19_sfbayarea/utils.py
@@ -1,0 +1,6 @@
+def friendly_county(county_id: str) -> str:
+    '''
+    Transform a county ID (e.g. "san_francisco") into a more human-friendly
+    name (e.g. "San Francisco").
+    '''
+    return county_id.replace('_', ' ').title()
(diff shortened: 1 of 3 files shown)


rickpr (approved):
Tried out the scrapers locally - output looks beautiful.

Good use of `click` - much nicer than just sending to stderr. Also `traceback` send to `stderr` by default, so this all looks good to me!

PR #126, opened 2020-09-12
HOTFIX: Alameda news page now has more `<br>`

Mr0grog:
The news scraper is currently failing for Alameda: https://github.com/sfbrigade/stop-covid19-sfbayarea/runs/1105106953?check_suite_focus=true

Previously, we only saw `<br>` elements on the Alameda news page as delimiters between the title and summary of a news item, or double `<br>`s as delimiters between news items. Now they're all over the place, including between the date and title. Since we previously assumed the `<br>` must be separating the title and summary, that meant we kicked into summary parsing mode without ever reading any title data. Luckily, that would cause the scraper to fail rather than output bad data (all news items MUST have titles), but that still means the scraper was broken.

The new approach here is to check to see whether we've gathered any title data yet when we hit a `<br>`. If not, keep looking for the title. Otherwise, treat it like before and start reading subsequent text as the summary.

### DIFF
--- a/covid19_sfbayarea/news/alameda.py
+++ b/covid19_sfbayarea/news/alameda.py
@@ -217,6 +217,7 @@ class ItemParser:
         Parse the news item that starts with the given HTML node. Raises
         `NotNews` if the HTML node isn't actually the start of a news item.
         """
+        logger.debug('Parsing node %s', self.start_node)
         self.state = self.START_STATE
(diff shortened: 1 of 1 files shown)


ldtcooper (approved):
This looks like a pretty straightforward change. It's working for me on Docker, and the results match what's on the page

PR #129, opened 2020-09-18
Fix Solano data scraper

elaguerta:
To run the scraper:
`$ python3 scraper_data.py solano`

Here is the main county dashboard: https://doitgis.maps.arcgis.com/apps/MapSeries/index.html?appid=055f81e9fe154da5860257e3f2489d67

Here is the output I received on Friday Sept 19:
https://gist.github.com/elaguerta/0be07f9b23af224ea882b88676866fa4

### DIFF
--- a/covid19_sfbayarea/data/solano.py
+++ b/covid19_sfbayarea/data/solano.py
@@ -17,6 +17,9 @@ data_url = "https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/C
 # data2_url looks like a join on Race/Eth and Age Group #TODO: Parse this nightmare
 data2_url = "https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/COVID19Surveypt2v3_view_3/FeatureServer/0/query"
 metadata_url = 'https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/COVID19Surveypt1v3_view/FeatureServer/0?f=pjson'
+
+# note, this is a link to just one dashboard, which is used to scrape the disclaimers.
+# The main public-facing interface with all dashboards is here: https://doitgis.maps.arcgis.com/apps/MapSeries/index.html?appid=055f81e9fe154da5860257e3f2489d67
 dashboard_url = 'https://doitgis.maps.arcgis.com/apps/opsdashboard/index.html#/d28335cd317a45cd84211cd290889c27'
 
 def get_county() -> Dict:
--- a/covid19_sfbayarea/data/solano.py
+++ b/covid19_sfbayarea/data/solano.py
@@ -208,9 +208,12 @@ def get_age_table(out: Dict) -> None:
     latest_day = datetime.fromtimestamp(latest_day_timestamp/1000,
                                         tz=timezone.utc).strftime('%m-%d-%Y')
 
-    # get all positive values for age group total cases on the latest day
+    # Get all positive values for age group total cases on the latest day.
+    # Skip the group "Total_AG", which represents the sum total of all groups.
     response2 = requests.get(data2_url, params={
-        'where': f"AG_Total_cases>0 AND Date_reported = '{latest_day}' ",
+        'where': f"""AG_Total_cases > 0
+                     AND Date_reported = '{latest_day}'
+                     AND Age_group <> 'Total_AG'""",
         'outFields': 'Age_group, AG_Total_cases, AG_deaths',
         'f': 'json'
     })
@@ -230,7 +233,7 @@ def get_age_table(out: Dict) -> None:
                         for entry in parsed2['features']]
 
     # Validate that we’ve got the right age groups.
-    if len(age_group_cases) != 5:
+    if len(age_group_cases) != 4:
         raise FormatError(f'Race_eth query did not return 5 groups. Results: {age_group_cases}')
 
     # save to the out dict
--- a/covid19_sfbayarea/data/solano.py
+++ b/covid19_sfbayarea/data/solano.py
@@ -17,7 +17,7 @@ data_url = "https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/C
 # data2_url looks like a join on Race/Eth and Age Group #TODO: Parse this nightmare
 data2_url = "https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/COVID19Surveypt2v3_view_3/FeatureServer/0/query"
 metadata_url = 'https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/COVID19Surveypt1v3_view/FeatureServer/0?f=pjson'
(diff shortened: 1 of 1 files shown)


elaguerta:
@benghancock If you have time over the next couple days, would you review this fix? Thanks!

elaguerta:
Note: I am taking Gender, Age Groups, and Race/Eth groups exactly as the county is reporting them, rather than mapping them to our common data model.

covid19_sfbayarea/data/solano.py
```diff
@@ -158,129 +142,105 @@ def get_notes() -> str:
...
-    # surface data from nested attributes dict
-    entries = [attr["attributes"] for attr in parsed['features']]
+    latest_day_timestamp = parsed['features'][0]['attributes']['Date_reported']
+    # translate timestamp  to a date string in format 'mm-dd-yyyy' to use in the query string
+    latest_day = datetime.fromtimestamp(latest_day_timestamp/1000, tz=timezone.utc).strftime('%m-%d-%Y')
 
-    if len(entries) != 4: # check that we have 4 entries, one for each expected age group
-        raise FormatError(
-            f"The source data structure has changed. Query did not return four age groups. Results: {entries}")
+    # get all positive values for race/ethnicity total cases on the latest day
```
Mr0grog, line 197:
```suggestion
    # get all positive values for age group total cases on the latest day
```

covid19_sfbayarea/data/solano.py
```diff
@@ -158,129 +142,105 @@ def get_notes() -> str:
...
-    This data used to have Age Group numbers. The Age_Group columns are still being shown in this source.
-    See the data at this map item: https://www.arcgis.com/home/webmap/viewer.html?url=https://services2.arcgis.com/SCn6czzcqKAFwdGU/ArcGIS/rest/services/COVID_19_survey_part_2_v2_public_view/FeatureServer/0&source=sd
-    The table view of the map item is a helpful reference.
-    """
-
-    # format query to get the latest 3 entries. This will get the latest Male and Female entries, plus one additional entry to check if there was
-    # an Unknown gender engry for the day
-    param_list = {'where': '0=0', 'outFields': '*',
-                  'orderByFields': 'date_reported DESC', 'resultRecordCount': '3', 'f': 'json'}
+     # filter for any days on which a Gender total cases numbner was reported
```
Mr0grog, line 222:
```suggestion
     # filter for any days on which a Gender total cases number was reported
```

covid19_sfbayarea/data/solano.py
```diff
@@ -158,129 +142,105 @@ def get_notes() -> str:
...
-            {"group": age_key, "raw_count": age_group_deaths})
+    age_group_deaths = { entry['attributes']['Age_group']: entry['attributes']['AG_deaths']  for entry in parsed2['features'] }
+    # save to the out dict
+    out["case_totals"]["age_group"] = age_group_cases
+    out["death_totals"]["age_group"] = age_group_deaths
 
-    out["case_totals"]["age_group"] = age_table_cases
-    out["death_totals"]["age_group"] = age_table_deaths
+    out["case_totals"]["age_group"] = age_group_cases
+    out["death_totals"]["age_group"] = age_group_deaths
```
Mr0grog, line 214: Looks like we are just doing the same thing twice here.

---

Also, this is outputting:

```js
"age_group": {
  "0_17_yrs": 661,
  "18_49_yrs": 3626,
  "50_64_yrs": 1160,
  "65_pl_yrs": 606,
  "Total_AG": 6054
},
```

But the format should be:

```js
"age_group": [
  {"group": "0_17_yrs", "count": 661},
  {"group": "18_49_yrs", "count": 3626},
  {"group": "50_64_yrs", "count": 1160},
  {"group": "65_pl_yrs", "count": 606},
  {"group": "Total_AG", "count": 6054}
]
```

covid19_sfbayarea/data/solano.py
```diff
@@ -158,129 +142,105 @@ def get_notes() -> str:
...
+    response2.raise_for_status()
+    parsed2 = response2.json()
+
+    gender_cases = { entry['attributes']['Gender']: entry['attributes']['G_Total_cases']  for entry in parsed2['features'] }
+    # A complete table will have at least 2 datapoints, as of 9/18/20. If were less than 2, raise an error.
+    if len(gender_cases) <2 :
+        raise FormatError( f'Gender query returned less than 2 groups. Results: {gender_cases}')
+
+    # save to the out dict
+    out["case_totals"]["gender"] = gender_cases
```
Mr0grog, line 243: This is outputting upper-case fields:

```js
"gender": {
  "Male": 2872,
  "Female": 3165
}
```

When they should be lower-case:

```js
"gender": {
  "male": 2872,
  "female": 3165
}
```

Mr0grog (commented):
Noted a couple issues inline, but otherwise this looks good to me.

Oof, looking at the source data, I hadn't realized how rough some of what's here is. It definitely seems like some deeper investigation is needed on some of this, but aside from the two minor issues noted inline, I feel good about this PR in terms of getting something in now.

> I suspect there is some double counting going on with the Race/Eth `Multirace/Other`, `Multirace`,  and `Other` categories

Yowza. Every one I spot-checked added up as “Multirace / Others” was not part of the total and was the only with duplicative data (so it’s as if you could just strip it out), which seems like a really suspicious pattern. But then the `RE_Pct_of_total` column (“Percent of total (all cases) for this R/E”) doesn’t add up in a similar way at all (in fact, it doesn’t track with the case numbers vs. the race/ethnicity's percentage of county population at all). I’m sure you were looking at the same things, but ugh. 

I think you should make an issue to investigate further, then you or anyone who wants to contribute can track learnings (or communications with the county?) in the issue thread.

---

I was also struggling with `DBC_total_number_of_specimens_collected` vs. `cumulative_cases`. If I’m reading the [data glossary](http://solanocounty.com/civicax/filebank/blobdload.aspx?BlobID=31926) right (my interpretation is questionable, plus some of the fields mentioned here aren’t actually in the data, sooooo… idunno), then:

- `DBC_total_number_of_specimens_collected` = “number of positive cases from tests that were run on samples taken this date” (The column name really does not imply what the data dictionary is saying here.)
- `cumulative_cases` = “number of cases we knew about as of this date”

IIRC from earlier reviews, we have some counties where the data is more like one of these and some where it’s more like the other. I don’t think there’s any reason to do differently from what you’re doing already for now, but longer term (i.e. *after* merging this), I’m wondering if we should double-check which of these is available and which we are using for each county, and officially bias towards one or the other when both are available, and also figure out how to message when we are showing different methods for county A vs. county B. (Apologies if you’ve already done that work somewhere and I missed it!)

---

Do you have any idea why the age groups don’t appear to add up to the total in the source data? (Or, at least, both dates I spot-checked appear to be off by 1.) Regardless, it seems like an issue in the source data, not your code.

Mr0grog:
Hey @elaguerta, just checking in — are you planning to finish out this PR by fixing the two output formatting issues, or do you need someone else to help push this across the finish line?

Mr0grog:
OK, I rebased this to get the schema tests from #130, which flagged both the issues I did above, so we’ve got some validation of the fixes here. Also did a little tiny bit of syntax cleanup while fixing the schema/formatting issues.

PR #130, opened 2020-09-19
Add tests to validate output format

Mr0grog:
It turns out several of our scrapers don't quite conform to the data model. We should probably have some classes that model the data and can enforce some formatting and validation, but it's also good to test that the final result adheres to a schema, which is what this does.

Because it's tough to mock responses in Selenium, these tests run against live county websites right now. That can make them pretty slow or intense to run, so this class of tests is disabled by default and you have to opt-in to run them (in CI, we opt-in). To do so, set the `LIVE_TESTS` environment variable to `*` to run all live tests, or set it to a comma-separated list of county names to run just those counties:

    # Run live tests against all county websites.
    $ LIVE_TESTS='*' python -m pytest -v .

    # Run live tests against only San Francisco and Sonoma counties.
    $ LIVE_TESTS='san_francisco,sonoma' python -m pytest -v .

Finally, **these tests skip any county that raises an error while scraping.** Because changes on county websites frequently cause scrapers to fail, we don’t want that to create flaky, failure-prone tests here. (For example, if Alameda and Solano are both failing, we don’t want to be bothered with Alameda errors on the tests for the PR that fixes Solano.)

The validations use [Voluptuous](https://github.com/alecthomas/voluptuous), which is a reasonably nice and concise language for writing schemas. A more verbose, but more standard alternative might be to use JSONSchema.

Tests will now output useful information about what’s wrong with the output data format. For example, in San Mateo:

```
________________ test_scraper_output_format_is_valid[san_mateo] ________________

...
(Lines from test here)
...

E       AssertionError: assert failed due to validation error(s):
E         - case_totals.gender: expected a dictionary (actual: [{'female': 4682}, {'male': 4646}, {'other': 1}, {'unknown': 3}])
E         - case_totals.age_group.0.count: extra keys not allowed (actual: 507)
E         - case_totals.age_group.0.raw_count: required key not provided
E         - death_totals.gender: expected a dictionary (actual: [{'female': 71}, {'male': 71}])
E         - death_totals.age_group.0.count: extra keys not allowed (actual: 0)
E         - death_totals.age_group.0.raw_count: required key not provided
```

Or Marin:

```
__________________ test_scraper_output_format_is_valid[marin] __________________

...
(Lines from test here)
...

E       assert failed due to validation error(s):
E         - update_time: does not match regular expression (actual: '2020-09-19T08:53:43.142176')
E         - meta_from_source: expected str (actual: (['COVID-19 testing and case data are reported as timely, accurately, and...
```

For a complete example, see this test run: https://app.circleci.com/pipelines/github/sfbrigade/data-covid19-sfbayarea/385/workflows/a57acb5d-6858-47b6-b84b-f0b8b5578b4f/jobs/549

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,9 @@
 version: 2.1
 
-jobs:
-  lint:
-    docker:
-      - image: circleci/python:3.7
+commands:
+  setup_dependencies:
+    description: "Set Up Dependencies"
     steps:
-      - checkout
       - restore_cache:
           keys:
             - data-covid-19-sfbayarea-v1-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
@@ -27,6 +25,14 @@ jobs:
           key: data-covid-19-sfbayarea-v1-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
           paths:
             - env
+
+jobs:
+  lint:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - checkout
+      - setup_dependencies
 
       - run:
           name: Pyflakes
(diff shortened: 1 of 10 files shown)


Mr0grog:
Whoops, I clearly didn’t test that last commit fully! It read nicer, but there’s no way to use `strptime` strings to represent all the ways a full datetime can be represented in ISO 8601. (We could fix that by restricting the time precision in all our datetimes, but it seemed better to change this validation for now.) It’s fixed now.

ldtcooper (approved):
Runs just fine in docker on Windows. Looks like I have some work to do on Sonoma, but the test themselves looks good.

rickpr (approved):
Awesome!!

Issue #132, opened 2020-09-22
CDS data has not been updated since 8/31/2020


kengoy:
Since 8/31/2020, the CDS site has not been updating data. Seems the public site is busted but the data is still updating and available directly from the updated link below. Keep watching the issue: https://github.com/covidatlas/li/issues/606

Mr0grog:
Fixed in #133.

PR #133, opened 2020-09-22
workaround to update the CDS data.

kengoy:
Workaround for #132. Seems the CDS public site is currently busted, but the data is still updating and available directly from s3 (https://liproduction-reportsbucket-bhk8fnhv1s76.s3-us-west-1.amazonaws.com/v1/latest/timeseries-byLocation.json)
Not the prettiest URL ... but it works! At least we can keep our site being updated.


Mr0grog (approved):
Looks like a good interim fix to me. 👍

PR #136, opened 2020-09-26
Fix date prefix removal in San Mateo news

### LINKED ISSUE #134: San Mateo news titles are prefixed with a date
San Mateo’s news items have their title prefixed with a date, which is confusing and redundant since we already display the date before the title for news items anyway:

Screen Shot 2020-09-24 at 10 00 37 PM

We currently have some code that used to remove these:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/238ef6d05ce45f2c7f08a3c39e716f12071cda57/covid19_sfbayarea/news/san_mateo.py#L19-L25

and:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/238ef6d05ce45f2c7f08a3c39e716f12071cda57/covid19_sfbayarea/news/san_mateo.py#L96-L97

but the format must have changed slightly so that regular expression no longer works.

If you run `python screaper_news.py san_mateo`, you now get entries like:

```js
    {
      "id": "12701 at https://cmo.smcgov.org",
      "url": "https://cmo.smcgov.org/press-release/sept-24-2020-safely-join-vote-center-team-new-health-officer-statement-testing",
      "title": "Sept. 24, 2020: Safely Join the Vote Center Team; New Health Officer Statement; Testing Opportunities: Update on County Response to COVID-19",
      "date_published": "2020-09-24T20:44:03Z"
    }
```

Mr0grog:
Fixes #134.

Many news items have their title prefixed with a date in San Mateo, which can be a bit confusing and redundant, especially when we normally display the date right before the title at https://panda.baybrigades.org.

Screen Shot 2020-09-24 at 10 00 37 PM

To alleviate this, we attempt to remove the date at the start of the title. However, the county recently changed its date format to include abbreviations, which we didn't previously account for. It also turns out that dates are getting formatted differently in Spanish (Chinese articles seem to have no date prefix, and I don't have examples of any other languages to test).

Additionally, it turns out news items with summaries have a similar date prefix on the summary, too! We now remove it as well.

Finally, I added some tests for the San Mateo scraper here. Obviously these won’t help us if the San Mateo source data changes, so these work more to help make sure nothing gets accidentally screwed up when someone is fixing another bug or refactoring something. (And any time the format of San Mateo’s source data changes, the mocks in these tests will also need to be updated.)

### DIFF
--- a/tests/news/san_mateo_test.py
+++ b/tests/news/san_mateo_test.py
@@ -1,0 +1,109 @@
+from covid19_sfbayarea.news.san_mateo import SanMateoNews
+from datetime import datetime, timezone
+from unittest.mock import patch
+
+
+def test_parses_news_data() -> None:
+    mock_rss = """
+    <rss version="2.0" xmlns:dc="http://purl.org/dc/elements/1.1/">
+        <channel>
+            <item>
+                <title>Sept. 25, 2020: Health Officer Order Prohibits Removal</title>
+                <link>https://cmo.smcgov.org/press-release/sept-25-2020-health-officer-order-prohibits-removal-fire-debris-burn-sites-pending</link>
+                <description><![CDATA[Sept. 25, 2020San Mateo County Health Officer Dr. Scott Morrow has issued a health order.]]></description>
+                <pubDate>Fri, 25 Sep 2020 22:05:08 +0000</pubDate>
+                <dc:creator>mwilson</dc:creator>
+                <guid isPermaLink="false">12711 at https://cmo.smcgov.org</guid>
+            </item>
+        </channel>
+    </rss>
+    """.encode('utf-8')
+
+    scraper = SanMateoNews()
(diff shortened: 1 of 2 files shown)


tests/news/san_mateo_test.py
```diff
@@ -0,0 +1,109 @@
+from covid19_sfbayarea.news.san_mateo import SanMateoNews
```
rickpr, line 1: Hey this is awesome! Great example of how to unit test something that I thought was near impossible to unit test.

rickpr (approved):
LGTM - nice way of documenting the regexes, it makes it easier not to get lost.

Mr0grog:
> nice way of documenting the regexes

Yeah! I’d seen these before in Perl scripts, but only recently discovered the `re.VERBOSE` flag that lets you do it in Python. Really nice.

PR #138, opened 2020-09-30
HOTFIX: Allow special spaces in more Contra Costa strings

Mr0grog:
In this recent news update, you’ll notice that all the Contra Costa news from this month just disappeared: https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/442/files

We already had some relatively complex space matching in some parts of Contra Costa regular expressions, but now they have dropped other complex space characters in new places that is causing the scraper to miss entire months' worth of news. To handle it, I’ve basically just made our headings and title matchers accept the more complex space character set anywhere they were previously using `\s`. 😩

You can test this by running:

```sh
# If you use the shell script:
$ ./run_scraper_news.sh contra_costa

# Or if you manage virtualenvs manually:
$ python scraper_news.py contra_costa
```

And make sure there are news items for September, and that the titles of most news items generally seem correct.

### DIFF
--- a/covid19_sfbayarea/news/contra_costa.py
+++ b/covid19_sfbayarea/news/contra_costa.py
@@ -23,14 +23,18 @@ MONTHS = (
     'december',
 )
 
+# This page uses all manner of complex space characters like em squares,
+# zero-width spaces, etc. that do not get matched by `\s` in regexes. To handle
+# that, we replace `\s` is some expressions with this more permissive character
+# class that includes many of those other kinds of spaces.
+PERMISSIVE_SPACE = r'[\s\u2000-\u200d]'
+
 # There's not a great way to identify a news item listing on the page (and very
 # few CSS classes or attributes in general), so we find news items by looking
 # for the month-based headings and finding all the <li> tags after them.
 MONTH_HEADING_PATTERN = re.compile(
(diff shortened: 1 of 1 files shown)


benghancock (approved):
This looks like it's working! I took a look through the output and compared against the site and it seems complete. I did notice that it doesn't appear as though we're grabbing anything earlier than Aug. 31, but I haven't done enough testing yet to figure out why that is. Is that by design?

Mr0grog:
> it doesn't appear as though we're grabbing anything earlier than Aug. 31, but I haven't done enough testing yet to figure out why that is. Is that by design?

Yep! You can adjust using the from `--from` option on the command line: https://github.com/sfbrigade/data-covid19-sfbayarea/blob/cc89fe4b9bb346db10134ed09a6fe9e7cc88e778/scraper_news.py#L66-L69

Mr0grog:
Thanks for the speedy review! 🙇

benghancock:
> Yep! You can adjust using the from `--from` option on the command line:

D'oh! Of course. I didn't spot the default value in there. Makes good sense.

benghancock:
> Thanks for the speedy review! bow

For sure! :+1: Glad to be able to return the favor!

PR #139, opened 2020-10-08
Fix Alameda data scraper

rickpr:


rickpr:
There is some refactoring work to be done here - there is duplicate code both within the scraper and between this scraper and San Mateo. This effort here is to get this working first and I'll follow up with the refactor.

Mr0grog (changes requested):
This is huge! But I can see the Power BI stuff is pretty complex. Definitely lots to abstract between this and San Mateo later. Given that, I tried to focus mostly on anything that seemed to be problematic in the output, rather than anything about specifically about the code.

There’s a few funky bits, but mostly this is awesome! I can see it was a big effort. 💪

covid19_sfbayarea/data/alameda/cases_by_gender.py
```diff
@@ -0,0 +1,27 @@
...
+        return [
+            {
+                'Column': self._column_expression(self.property),
+                'Name': f'{self.name}.{self.property}'
+            },
+            {
+                'Measure': self._column_expression(property),
+                'Name': f'{self.name}.{property}'
+            },
+       ]
```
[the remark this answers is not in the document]
Mr0grog: Ahhhhhhhh, I see what’s different! The dashboard’s query has a `Where` section that explicitly selects out `Male`, `Female`, and `Unknown` (also, only rows with a value `> 10`), and we don’t have a `Where` section at all:

```js
{
    "version": "1.0.0",
    "queries": [
        {
            "Query": {
                "Commands": [
                    {
                        "SemanticQueryDataShapeCommand": {
                            "Query": {
                                "Version": 2,
                                "From": [
                                    {
                                        "Name": "v",
                                        "Entity": "V_Combined_data",
                                        "Type": 0
                                    }
                                ],
                                "Select": [
                                    {
                                        "Column": {
                                            "Expression": {
                                                "SourceRef": {
                                                    "Source": "v"
                                                }
                                            },
                                            "Property": "Gender"
                                        },
                                        "Name": "V_Combined_data.Gender"
                                    },
                                    {
                                        "Measure": {
                                            "Expression": {
                                                "SourceRef": {
                                                    "Source": "v"
                                                }
                                            },
                                            "Property": "NumberOfCases"
                                        },
                                        "Name": "V_Combined_data.NumberOfCases"
                                    }
                                ],
                                "Where": [
                                    {
                                        "Condition": {
                                            "In": {
                                                "Expressions": [
                                                    {
                                                        "Column": {
                                                            "Expression": {
                                                                "SourceRef": {
                                                                    "Source": "v"
                                                                }
                                                            },
                                                            "Property": "Gender"
                                                        }
                                                    }
                                                ],
                                                "Values": [
                                                    [
                                                        {
                                                            "Literal": {
                                                                "Value": "'Female'"
                                                            }
                                                        }
                                                    ],
                                                    [
                                                        {
                                                            "Literal": {
                                                                "Value": "'Male'"
                                                            }
                                                        }
                                                    ],
                                                    [
                                                        {
                                                            "Literal": {
                                                                "Value": "'Unknown'"
                                                            }
                                                        }
                                                    ]
                                                ]
                                            }
                                        }
                                    },
                                    {
                                        "Condition": {
                                            "Comparison": {
                                                "ComparisonKind": 2,
                                                "Left": {
                                                    "Measure": {
                                                        "Expression": {
                                                            "SourceRef": {
                                                                "Source": "v"
                                                            }
                                                        },
                                                        "Property": "NumberOfCases"
                                                    }
                                                },
                                                "Right": {
                                                    "Literal": {
                                                        "Value": "10L"
                                                    }
                                                }
                                            }
                                        },
                                        "Target": [
                                            {
                                                "Column": {
                                                    "Expression": {
                                                        "SourceRef": {
                                                            "Source": "v"
                                                        }
                                                    },
                                                    "Property": "Gender"
                                                }
                                            }
                                        ]
                                    }
                                ],
```

Where this *would* go in our code:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/32f8fc03e6b8ecb5ef680cbecb86f2db8b4675a3/covid19_sfbayarea/data/alameda/power_bi_querier.py#L52-L63

Part of me says we should follow whatever the dashboard is doing, but I worry we won’t know when to change if they ever add other genders. Maybe it’s best just to either:

- Specifically exclude `D` (since it seems like it’s probably bad data).
- Strip out values less than 10 (I think some of our other data sources have that built-in, and the Alameda dashboard seems to be doing it in all the categorizations for cases and for deaths), since those represent potential privacy issues anyway.

Or even do both, I guess. We could include `< 10` in the query but output something in our data so the category is there, but the front-end can display it as “unknown” or “too few” or whatever.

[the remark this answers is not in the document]
rickpr: I was thinking the `Where` clause was unnecessary as it would only serve to limit the data. I'm going to have a look at stripping these and maybe include the where clause instead as it is what the dashboard is doing. I hadn't thought about privacy issues for small numbers of cases.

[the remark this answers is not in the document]
Mr0grog: Heh, I would have made the same assumption until I saw the results! Plus not filtering seems more future-proof (hence why I think it might be better to filter *out* `D` instead of copying the dashboard's query and only asking for specific fields). I admit I’m pretty curious what `D` and the other complex field are, though. 😛

Re: privacy issues, it probably only jumped out at me because one of the other county scrapers used to have special code to handle it (that county returned integers most of the time, but had the string "< 10" when there were too few). Not showing a number for < 10 seems like a good heuristic to follow, but I don't know that we do it explicitly anywhere else — we just count on the fact that the county is only exposing values that have already been filtered. So I think(?) this may be a unique case.

covid19_sfbayarea/data/alameda/meta.py
```diff
@@ -0,0 +1,44 @@
...
+        try:
+            url = ''.join([
+                'https://wabi-us-gov-iowa-api.analysis.usgovcloudapi.net/public/reports/',
+                PowerBiQuerier.DEFAULT_POWERBI_RESOURCE_KEY,
+                '/modelsAndExploration?preferReadOnlySession=true'
+            ])
+            response = get(url, headers = { 'X-PowerBI-ResourceKey': PowerBiQuerier.DEFAULT_POWERBI_RESOURCE_KEY })
+            return self._extract_meta(response.json())[2:] # First two characters are ': '
+        except:
+            return """
```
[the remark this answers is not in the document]
rickpr: Not really other than the fact that I didn't expect it to change. This is a good point though - if it changes we shouldn't fall back to an old meta.

covid19_sfbayarea/data/alameda/time_series_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        self._assert_total_and_percent_cases_count_matches(total_tests, percent_positive_tests)
+
+        results = [{
+            'date': self._timestamp_to_date(timestamp),
+            'tests': total_tests[timestamp],
+            'pending': -1, # we don't have data for this
+            'cumul_pend': -1, # no data for this
+            **self._positive_and_negative_tests(total_tests[timestamp], percent_positive_tests[timestamp])
+        } for timestamp in total_tests.keys()]
+        self._add_cumulative_data(results)
```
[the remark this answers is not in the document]
rickpr: Added to the meta and dropped these. I'm looking for a way to get this data, but it isn't exposed on the dashboard, so will require some experimentation.

rickpr:
I believe I've addressed all the requested changes. I also added a bump to Python 3.8 due to this: https://github.com/sfbrigade/data-covid19-sfbayarea/pull/139/files#diff-8c4098ea3665049c04627d252f3e6c0df96405eeecc13838b8dd9029b736c46dR19

rickpr:
Also, output from a local run is here: https://gist.github.com/rickpr/82cddf12332c72b16b6483cc946ba230

Mr0grog:
Oooh, using perhaps the fanciest of new (ish) Python features!

covid19_sfbayarea/data/alameda/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+        'source_url': LANDING_PAGE,
+        'meta_from_source': Meta().get_data(),
+        'meta_from_baypd': """
+            See power_bi_scraper.py for methods.
+            Alameda does not provide a timestamp for their last dataset update,
+            so BayPD uses midnight of the latest day in the cases timeseries as a proxy.
+
+            The test cases are on a rolling seven-day average and do not represent the
+            exact number of cases on any given day.
+         """,
```
Mr0grog, line 43: Minor nit: this line is over-indented by one space.

covid19_sfbayarea/data/alameda/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    last_updated = most_recent_case_time(data)
+    data.update({ 'update_time': last_updated.isoformat() })
+    return data
+
+def most_recent_case_time(data: Dict[str, Any]) -> datetime:
+    most_recent_cases = cast(Dict[str, str], dig(data, ['series', 'cases', -1]))
+    pacific_time = tz.gettz('America/Los_Angeles')
+    # Offset by 8 hours to ensure the correct day is shown
+    start_of_day_pacific = datetime.strptime(most_recent_cases['date'] + '-8', '%Y-%m-%d-%H')
+    return start_of_day_pacific.astimezone(pacific_time)
```
Mr0grog, line 69: A more correct way to do this that reliably sets timezone without messing about with the hours is:

```suggestion
    pacific_time = tz.gettz('America/Los_Angeles')
    return (datetime.strptime(most_recent_cases['date'], '%Y-%m-%d')
            .replace(tzinfo=pacific_time))
```

However, we already have a helper function to do all this that might be worth using at `news.utils.parse_datetime`:

```py
    return parse_datetime(most_recent_cases['date'])
```

If you use that, you should probably move `parse_datetime` and `PACIFIC_TIME` up to the top-level `utils` module.

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/d1a4458526b38a2040b0d10be54fe14e5143c9e5/covid19_sfbayarea/news/utils.py#L114-L133

covid19_sfbayarea/data/alameda/cases_by_age.py
```diff
@@ -0,0 +1,26 @@
...
+class CasesByAge(PowerBiQuerier):
+    def __init__(self) -> None:
+        self.source = 'v'
+        self.name = 'V_Combined_data'
+        self.property = 'AgeGroup'
+        super().__init__()
+
+    def _parse_data(self, response_json: Dict[str, List]) -> List[Dict[str, int]]:
+        results = super()._parse_data(response_json)
+        return [ { 'group': group, 'raw_count': count } for group, count in results if group != 'Unknown Age' ]
```
Mr0grog, line 13: I’ve lost track and can’t check since you rebased; were you dropping the “Unknown Age” group before? Should we be doing that/is there a value there we should show?

covid19_sfbayarea/data/alameda/deaths_by_age.py
```diff
@@ -0,0 +1,26 @@
...
+class DeathsByAge(PowerBiQuerier):
+    def __init__(self) -> None:
+        self.source = 'v'
+        self.name = 'V_Combined_data'
+        self.property = 'AgeGroup'
+        super().__init__()
+
+    def _parse_data(self, response_json: Dict[str, List]) -> List[Dict[str, int]]:
+        results = super()._parse_data(response_json)
+        return [ { 'group': group, 'raw_count': count } for group, count in results if group != 'Unknown Age' ]
```
Mr0grog, line 13: Same as for cases w/r/t “Unknown Age” here.

covid19_sfbayarea/data/alameda/utils.py
```diff
@@ -0,0 +1,9 @@
+from functools import reduce
+from typing import Any, Dict, List
+
+def dig(items: Dict[str, Any], json_path: List[Any]) -> Any:
+    try:
+        return reduce(lambda subitem, next_step: subitem[next_step], json_path, items)
+    except (KeyError, TypeError, IndexError) as err:
+        print('Error reading returned JSON, check path: ', err)
+        raise(err)
```
Mr0grog, line 9: Not a big deal, but this is generic enough that you might as well move it up one or two levels to `covid19_sfbayarea.data.utils` or `covid19_sfbayarea.utils`.

Mr0grog (commented):
Added a small handful of comments inline. The only thing that’s a showstopper is that, if we’re updating to Python 3.8, the README absolutely needs to be updated to reflect that.

Everything else is a minor issue or suggestion that might improve things, but isn’t really critical for merging.

covid19_sfbayarea/data/alameda/utils.py
Mr0grog: Oh, just noticed a minor typing issue: if `items` is always keyed by strings (`Dict[str, Any]`), `json_path` should probably be restricted to a list of strings (`List[str]`).

rickpr: Oh this is a good point - `items` should be a `dict` but can contain a `list` at a deeper level, so `ints` are appropriate, so it should maybe be `List[Union[int, str]]`.

covid19_sfbayarea/data/alameda/cases_by_age.py
rickpr: Yes - I had been dropping this before. There is a value there, but I don't know that it fits in with our frontend dashboard. Should we preserve it?

rickpr:
README update here: https://github.com/sfbrigade/data-covid19-sfbayarea/pull/139/commits/8fc58a8bd1687179afbef1081e9f68972524f325

covid19_sfbayarea/data/alameda/__init__.py
rickpr: Cool, using the helper function here: https://github.com/sfbrigade/data-covid19-sfbayarea/pull/139/commits/c6aac13df6c0c8c7ea484c1fae5a94bfc65d65a3

covid19_sfbayarea/data/alameda/utils.py
rickpr: Moved it to upper `utils` and I think this is correct typing now: https://github.com/sfbrigade/data-covid19-sfbayarea/pull/139/commits/e7b3520e0feda1899a7110b3259108a8103c96b8

covid19_sfbayarea/data/alameda/cases_by_age.py
Mr0grog: My tendency is generally towards keeping it in unless we know it’s problematic to have (e.g. the issue with having per-ethnicity counts and totals together in the first review). BUT the right answer here is probably to ask the front-end folks when you’re not sure.

That said, it doesn’t seem like the decision should block getting this in. Just don’t let that be a reason to drop and not address it after. :)

Mr0grog (approved):
Looks ship-able to me! :shipit:

covid19_sfbayarea/data/alameda/cases_by_age.py
rickpr: Assigning myself this as an issue so I can sync with the frontend team at hack night: https://github.com/sfbrigade/data-covid19-sfbayarea/issues/147

PR #141, opened 2020-10-10
HOTFIX: Handle messy links in Alameda news titles

### LINKED ISSUE #140: Alameda news titles are cut off when links are in the middle of the title
The Alameda news page’s formatting continues to be kind of janky and continually introduce new quirks. We now have a news item where only a single letter in the title is a link, and that causes us to cut off the title where the link starts.

This page content for a news item on 2020-09-28:

Screen Shot 2020-10-09 at 9 31 01 AM

And outputs the following news item in the parsed result:

```js
{
  "id": "https://covid-19.acgov.org/covid19-assets/docs/press/joint-statement-2020.09.28.pdf",
  "url": "https://covid-19.acgov.org/covid19-assets/docs/press/joint-statement-2020.09.28.pdf",
  "title": "Guidance on How to Celebrate Halloween and D",
  "date_published": "2020-09-28T00:00:00-07:00"
}
```

You can see that the title cuts off before the “í” in “Día de los Muertos,” which is the link.

(See also this automatic PR that was created with the scraper: https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/484#discussion_r502546367)

## To Reproduce:

Run the news scraper for Alameda and look for the item described above:

```sh
$ python scraper_news.py alameda
```

## A First Guess at the Cause and Suggested Fix

I’m pretty sure the problem here is that, as we walk our way through the content of a news item on the page, we hit the link. Because it only has one word, we treat it like a language link and stop parsing it as part of the title:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/d0337819fc3bc8773ff5195fdd64564e41c24b87/covid19_sfbayarea/news/alameda.py#L324-L330

(I used one word as the condition for detecting a language because I was worried about knowing all the possible language names that could be used, and because this page is so riddled with errors, I wouldn’t be surprised if there were typos in the language names.)

So this can probably just be fixed by making the conditions in `is_language_link()` more narrow.

Mr0grog:
The Alameda news page now has links that take up only part of the title. The parser sometimes identifies these as links to language-specific versions of the news item, which causes it to stop parsing the title and start looking for language links. However, these links aren't language links, and we need to continue parsing the title as we move through them.

For example, the Alameda news page currently has this entry:

Screen Shot 2020-10-09 at 9 31 01 AM

…which was causing us to output the title:

> Guidance on How to Celebrate Halloween and D

…instead of:

> Guidance on How to Celebrate Halloween and Día de los Muertos Safely

This fixes the issue by being *much* more specific about what we expect for the text of a language link. This may have issues if there are ever typos in the language names, but seems to work well enough.

Fixes #140.

### DIFF
--- a/covid19_sfbayarea/news/alameda.py
+++ b/covid19_sfbayarea/news/alameda.py
@@ -13,6 +13,29 @@ from .utils import get_base_url, parse_datetime
 logger = getLogger(__name__)
 
 
+# Maps the various forms of language names in the page's text to ISO language
+# codes so we can determine the language of a link.
+# TODO: Actually use these to handle more than just English links.
+LANGUAGES = {
+    'english': 'en',
+    'spanish': 'es',
+    'español': 'es',
+    'chinese': 'zh_HANS',
+    'chinese (simplified)': 'zh_HANS',   # 'zh_CN' is common, but less correct.
+    'chinese (traditional)': 'zh_HANT',  # 'zh_TW' is common, but less correct.
+    'korean': 'ko',
+    'vietnamese': 'vi',
+    'amharic': 'am',
+    'arabic': 'ar',
+    'farsi': 'fa',
+    'persian': 'fa',
+    'pashto': 'ps',
+    'urdu': 'ur',
+}
+
+LANGUAGE_NAMES = set(name.casefold() for name in LANGUAGES)
+
+
 def date_from_node_text(node: element.Tag) -> Optional[datetime]:
     """
     If an element contains a date as text, return the parsed date.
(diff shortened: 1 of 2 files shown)


covid19_sfbayarea/news/alameda.py
```diff
@@ -13,6 +13,29 @@
...
+    'chinese (traditional)': 'zh_HANT',  # 'zh_TW' is common, but less correct.
+    'korean': 'ko',
+    'vietnamese': 'vi',
+    'amharic': 'am',
+    'arabic': 'ar',
+    'farsi': 'fa',
+    'persian': 'fa',
+    'pashto': 'ps',
+    'urdu': 'ur',
+}
```
Mr0grog, line 34: This mapping isn’t strictly necessary at the moment, but we currently drop non-English links from the feed entirely. Having this mapping will allow someone to come through later a update the code to pick up non-English links and surface them in a meaningful way, whatever that might be.

Mr0grog:
Merging without review since this is a hotfix.

PR #143, opened 2020-10-12
HOTFIX: Handle new Contra Costa translation format

### LINKED ISSUE #142: Contra Costa news items with Spanish translations are getting skipped
The Contra Costa news page appears to have recently changed the format of how Spanish translations are displayed for some news items, and we are now missing those items entirely when scraping news. The items with the new translation style are list items (like the rest), but they have a sub-list of translations (looks like only Spanish, but I guess this makes for a good place to add others in the future). It looks like this:

Screen Shot 2020-10-12 at 1 24 28 PM

When running `scraper_news.py contra_costa`, the output doesn’t include the news item from October 10th shown above, but does include the others.

This is probably because we use the following regular expression to parse *the entire text* of the list item containing the news item:

https://github.com/sfbrigade/data-covid19-sfbayarea/blob/ada41d30f1b3742a475e7353d18997ac3c5b624c/covid19_sfbayarea/news/contra_costa.py#L41-L49

That regular expression looks for strings that *end* after the date, but items with the new style have a bunch more content (the sub-list of translations) afterward.

**The old style of translation is *also* still used, so we need to continue to keep supporting that.** For example:

Screen Shot 2020-10-12 at 1 31 40 PM

Mr0grog:
Contra Costa's news page now sometimes shows Spanish translations of news items in a new format (some items also use the old format). Items in the new format contain additional text beyond the main news heading, causing us to fail to extract news data from them (and therefore skip them as if they were not news items).

The actual change here is just a one-liner. The rest is adding tests. :)

Fixes #142.

You should be able to test this by making sure the news items described in #142 are correctly showing up in the resulting output when you run:

```sh
$ python scraper_news.py contra_costa
```

You can also just compare the output to the actual [Contra Costa News Page](https://www.coronavirus.cchealth.org/health-services-updates) and make sure all the items are listed and formatted correctly.

### DIFF
--- a/covid19_sfbayarea/news/contra_costa.py
+++ b/covid19_sfbayarea/news/contra_costa.py
@@ -45,7 +45,6 @@ ARTICLE_TITLE_PATTERN = re.compile(r'''
     (?:\s*\|\s*\w+)?  # An optional `| <language>` alternate language link
     \s*-\s*           # A `-` separator
     (\d+/\d+/\d+)     # The date in mm/dd/yyyy format
-    \s*$
 '''.replace(r'\s', PERMISSIVE_SPACE), re.VERBOSE)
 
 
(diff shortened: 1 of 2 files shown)


Mr0grog:
Going ahead and merging this, since it’s a hotfix, it’s been more than a day, and nobody has signed up to review.

PR #145, opened 2020-10-18
HOTFIX: Ensure news items are unique

Mr0grog:
Napa county's website is currently repeating each news item 33 times (!), so we are, too:

Screen Shot 2020-10-18 at 11 15 11 AM

While this isn't necessarily our fault, we could easily handle this by deduping news items when we add them to the feed.

You can test this by running:

```sh
$ ./run_scraper_news.sh napa
# Or:
$ python scraper_news.py napa
```

And making sure each item in the output is only shown once.


Mr0grog:
Going ahead and merging this to unblock pending news updates on the front-end, since it’s a hotfix.

PR #148, opened 2020-11-01
HOTFIX: Handle Alameda news with more nested nodes

Mr0grog:
Alameda news items used to follow a pretty standard pattern where none of the components are contained in other elements. It typically looked like:

```html
<strong>Date</strong> <a href="...">Title</a>
```

But now some parts of some news items are getting wrapped in `div` elements, causing us to have trouble iterating through all the parts and extracting the complete data. For example, a news item on the page now looks like this:

```html
<div><strong>Date</strong></div> <a href="...">Title</a>
```

Note this isn't really consistent across news items, so it's all pretty messy. In any event, the new structure causes a couple problems that get worked around in this change:

1. We have to find a containing element that we don't look beyond for news content (otherwise we might include other text on the page erroneously as part of the title or summary of the last news item). We used to do that by finding the parent of the element containing the date. That obviously doesn't work anymore.

2. We scan the element containing the date as a whole (rather than iterating through its contents), then skip forward to the next element to start reading the title. The way we did that before assumed the title was a sibling to the date element, which is no longer always true.

**Note there are some other news problems in Alameda, too! This change at least allows the Alameda news scraper to work, even though it won’t detect all news.**

## To Test

Run the news scraper and make sure it outputs news data, rather than failing with an error:

```sh
$ python scraper_news.py alameda
```

### DIFF
--- a/covid19_sfbayarea/news/alameda.py
+++ b/covid19_sfbayarea/news/alameda.py
@@ -245,6 +245,18 @@ class ItemParser:
         self.state = self.START_STATE
         node = self.start_node
         root = node.parent
+        # For some news items, the start tag is inside another element, like:
+        #   <div><strong>October 28, 2020</strong></div> News Title
+        # So we can't treat its parent as the root container of the whole news
+        # item. Instead, we look for the nearest ancestor with a class name and
+        # treat that as the root. It's a rough heuristic, but seems to work.
+        while root.get('class') is None:
+            if root.parent:
+                root = root.parent
+            else:
+                root = node.parent
+                break
+
         while node and root in node.parents:
             try:
                 self.before_node(node)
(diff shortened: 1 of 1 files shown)


rickpr (approved):
Tested it locally and error is fixed!

As a note - I had to manually specify the binary location when running the Firefox webdriver on Fedora. I don't know if this has to do with which version I am running (latest).

Mr0grog:
> I had to manually specify the binary location when running the Firefox webdriver on Fedora. I don't know if this has to do with which version I am running (latest).

You need to do this with any of our Selenium-based scrapers, though, right? I assume this is because of [the bug with `webdrivermanager` v0.8.0 with Geckodriver on Linux](https://github.com/rasjani/webdrivermanager/issues/35) (we should upgrade; I think that bug has been fixed, but we should also look at Playwright or Pyppeteer).

rickpr:
Thanks! I had forgotten about the `install.sh` and had picked up the binary manually. Using `0.9` of `webdrivermanager` solves the first issue, I am still running into this: https://github.com/mozilla/geckodriver/issues/1756

The reason is because the `firefox` in my path actually runs this shell script that does a bunch of things, not just launches firefox, which messes with the way `geckodriver` sends the commands. Looks like this has been fixed in nightly Mozilla already. Once that's merged into main Firefox, I'll open a PR to upgrade `webdrivermanager`.

Either way, it seems to work in our Docker container and GitHub actions, so I think it's something specific to my environment. I'll

Mr0grog:
Ooooof, fixed almost 3 months ago but not yet relased. :(

Issue #147, opened 2020-10-28
Alameda scraper "Unknown Age"


rickpr:
County or Counties: Alameda
Data source: https://covid-19.acgov.org/data.page
# Description
Age groups in Alameda county have an unknown group. Collect this data and sync with the frontend team to see how it would be handled.

See this comment for some context: https://github.com/sfbrigade/data-covid19-sfbayarea/pull/139#discussion_r513231363

PR #150, opened 2020-11-05
Do capture unknown age for Alameda

rickpr:
Per conversation with the team, we don't want to hide the Unknown Age. We already show this for San Mateo County, we should do it here too.

Closes https://github.com/sfbrigade/data-covid19-sfbayarea/issues/147


Mr0grog (approved):
Looks good to me! 👍

PR #152, opened 2020-11-05
HOTFIX: Separate tables for cases and deaths

Mr0grog:
The Marin scraper has been failing for a while — it turns out they separated cases and deaths into different charts/tables. This is pretty much the minimum fix to get it running.

Fixes #146.

## Testing

You can test this by running the Marin data scraper and making sure it outputs actual data instead of an error message:

```sh
$ python scraper_data.py marin
```

PR #153, opened 2020-11-05
Stop reloading the Marin dashboard repeatedly while parsing it

Mr0grog:
While working on #152, I noticed that the Marin scraper makes approximately 1 bajillion requests to the same URL (the Marin dashboard web page) — it generally starts a new web browser and then loads a clean copy of the page for each area that it wants to parse. That makes the scraper a little slow, but more importantly, puts us at risk of getting banned or rate limited if Marin ever adds spam/DOS prevention security tools to their site.

That can obviously be a little more efficient, so I went ahead and tried to slim that down by passing a webdriver instance around instead of just a URL. I probably went a bit overboard by wrapping the webdriver code in `MarinDashboardPage` class that does all the more abstract stuff like extracting and parsing chart data. I didn’t really change method signatures or logic much, though, so this change looks a lot bigger than it really is.

### DIFF
--- a/covid19_sfbayarea/data/marin.py
+++ b/covid19_sfbayarea/data/marin.py
@@ -77,8 +77,8 @@ def get_chart_data(url: str, chart_id: str) -> List[str]:
 def get_chart_meta(url: str, chart_ids: Dict[str, str]) -> str:
     """This method gets all the metadata underneath the data wrapper charts and the metadata at the top of the county dashboard."""
     # Some metadata strings are repeated, so use sets to dedupe them.
-    metadata: set = set()
-    chart_metadata: set = set()
+    metadata: List[str] = []
+    chart_metadata: List[str] = []
 
     with get_firefox() as driver:
         driver.implicitly_wait(30)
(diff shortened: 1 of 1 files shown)


covid19_sfbayarea/data/marin.py
```diff
@@ -27,115 +92,72 @@ def get_county() -> Dict:
...
+    if paragraphs:
+        # TODO: it's not clear why any of these are being removed, nor
+        # why they are not being replaced with an equivalent ASCII
+        # character or just a space (not having something else in their
+        # place results in joined up words, like "arealways")
+        # \u2014 = em dash
+        # \u00a0 = non-breaking space
+        # \u2019 = apostrophe/right single quote
+        metadata.extend(paragraph.text.replace("\u2014","").replace("\u00a0", "").replace("\u2019","")
+                        for paragraph in paragraphs)
```
Mr0grog, line 131: I *think* the right thing here is to leave the em-dashes and apostrophes in, but replace the non-breaking space with a normal space (although even that is debatable — in the news scraper, we generally err on the cautious side of not changing them in case it really is important that two words stay together).

Would love any feedback on this.

rickpr (approved):
Tested out locally and noticed two differences:
1. The scraping is much faster.
2. The ordering of the text in the meta is fixed, it was non-deterministic before.

Approved!

PR #157, opened 2020-11-11
HOTFIX: Add "Other, non-Hispanic" race to Sonoma

Mr0grog:
Sonoma county recently added a new race/ethnicity group ("Other, non-Hispanic"), so its scraper needs to be updated to account for it.

**To Test:**

Run `python scraper_data.py sonoma` and make sure it outputs data instead of an error.


rickpr (approved):
Reproduced the crash on master, this fixes it.

Approved!

Mr0grog:
Thanks! 🎉

Issue #110, opened 2020-08-03
San Francisco news does not include summary of each item


Mr0grog:
San Francisco’s news pages (https://sf.gov/news/topics/794) include summaries for each item, but the scraper does not currently pick them up.

The summary text should be included as the `summary` field on `NewsItem` instances created in the SF news scraper in `covid19_sfbayarea/news/san_francisco.py`: https://github.com/sfbrigade/data-covid19-sfbayarea/blob/06342ce29cf95b68c115cf0660ec904489e7f5b7/covid19_sfbayarea/news/san_francisco.py#L46-L62

PR #158, opened 2020-11-17
Add summary to San Francisco NewsItem

slim-builder:
Fix for issue #110

### DIFF
--- a/covid19_sfbayarea/news/san_francisco.py
+++ b/covid19_sfbayarea/news/san_francisco.py
@@ -5,7 +5,7 @@ from urllib.parse import urljoin
 from .base import NewsScraper
 from .errors import FormatError
 from .feed import NewsItem
-from .utils import get_base_url, HEADING_PATTERN
+from .utils import get_base_url, HEADING_PATTERN, normalize_whitespace
 
 
 class SanFranciscoNews(NewsScraper):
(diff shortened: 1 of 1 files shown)


covid19_sfbayarea/news/san_francisco.py
```diff
@@ -59,4 +59,6 @@ def parse_news_item(self, item: element.Tag, base_url: str) -> NewsItem:
         date_string = item.find('time')['datetime']
         date = dateutil.parser.parse(date_string)
 
-        return NewsItem(id=url, url=url, title=title, date_published=date)
+        summary = item.find('div').find('div').find('div').get_text(strip=True)
```
Mr0grog, line 62: Two quick thoughts here:

1. Would you please use `news.utils.normalize_whitespace(element.get_text())` instead of `element.get_text(strip=True)`? I realized a while back that `get_text(strip=True)` is actually a bad pattern in most cases, but I guess I failed to fix it everywhere (totally understand if you were following the example of `title = ...` above, which is my fault for not updating).

2. Is there any better sort of selector we can build from classes, or maybe take the next sibling after the title heading, rather than just going 3 divs deep? This strikes me as pretty error-prone.

Mr0grog (changes requested):
Hey @slim-builder, thanks so much for this! I’m sorry none of us here saw it for so long; if you post another PR and it doesn’t get a reasonably quick review, please post on the SF Brigade Slack to make sure someone’s at least seen it.

Mr0grog:
@slim-builder is this ready for a re-review/merge?

Mr0grog (approved):
Looks wonderful, thank you! 👍 👍 👍

PR #159, opened 2020-12-17
HOTFIX: Clean up a variety of errors in data scrapers

Mr0grog:
It turns out the scrapers have been failing for a while because of a stray `print()` call, and that failure masked a series of other, further failures from a variety of counties. This attempts to fix them all:

- Clean up a couple stray `print()` calls (this should be all of them). One should not have existed at all, and the other I’ve replaced with a more detailed exception that I think actually does the job better.

- Clean up what appears to be unintentional spaces in Marin county’s table headers.

- Support new genders in San Francisco’s data.

- Support new age groups in San Francisco’s data, and also switch to human-friendly, county defined age groups (we decided to do this for all counties a while back, and this error seemed like a good time to go ahead and do it for SF rather then defining a whole new list of semi-standard age group keys).

- Fix renamed fields in San Mateo’s API. This also adds nicer error handling for PowerBI queries.

### DIFF
--- a/covid19_sfbayarea/data/san_mateo/power_bi_querier.py
+++ b/covid19_sfbayarea/data/san_mateo/power_bi_querier.py
@@ -1,7 +1,9 @@
 import json
 from requests import post
 from typing import Any, Dict, List, Union
-from covid19_sfbayarea.utils import dig
+from ...utils import dig
+from ...errors import PowerBiQueryError
+
 
 class PowerBiQuerier:
     BASE_URI = 'https://wabi-us-gov-iowa-api.analysis.usgovcloudapi.net/public/reports/querydata?synchronous=true'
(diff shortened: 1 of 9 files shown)


Mr0grog:
@rickpr I made some PowerBI changes here you might want to include in #151. There’s some slightly nicer error reporting and I had to update two property names in San Mateo’s data.

covid19_sfbayarea/data/san_mateo/power_bi_querier.py
```diff
@@ -1,7 +1,9 @@
 import json
 from requests import post
 from typing import Any, Dict, List, Union
-from covid19_sfbayarea.utils import dig
+from ...utils import dig
```
rickpr, line 4: Hey cool, I didn't know how to do this.

rickpr (approved):
LGTM - tested locally and the data scraper output looks sane to me.

I will incorporate these changes in #151, I have been out the last couple weeks so haven't worked on it, this is a great inspiration to get back to work on it, I will have an update next week!

PR #160, opened 2020-12-18
HOTFIX: Handle strings in Alameda %-positive tests

Mr0grog:
For some days in Alameda, we get strings instead of numbers in the percent positive tests timeseries. This fix is not particularly generic, but solves the immediate problem. It looks like PowerBI may try to preserve precision by sending strings when float64 cannot accurately represent a value. We'll have to think about more generic ways to account for this.

Example excerpt from the data:

```json
{
  "results": [
    {
      "result": {
        "data": {
          "dsr": {
            "DS": [
              {
                "PH": [
                  {
                    "DM0": [
                      {"C": [1607644800000, 6.916541]},
                      {"C": [1607731200000, 7.157232]},
                      {"C": [1607817600000, "7.2322939999999996"]},  // Not a number!
                      {"C": [1607904000000, 7.288172]},
                      {"C": [1607990400000, 7.475967]}

### DIFF
--- a/covid19_sfbayarea/data/alameda/time_series_tests_percent.py
+++ b/covid19_sfbayarea/data/alameda/time_series_tests_percent.py
@@ -29,3 +29,16 @@ class TimeSeriesTestsPercent(PowerBiQuerier):
             },
             'Version': 1
         }
+
+    def get_data(self) -> List:
+        # The values in this dataset can sometimes be strings. It looks like
+        # PowerBI probably sends strings for values that cannot be represented
+        # by a float64 (since JSON only supports 64-bit floats for numbers).
+        # For example, '7.2322939999999996' gets sent as a string; the closest
+        # float64 representation is '7.232294'.
+        #
+        # For our purposes, the loss of precision here is OK, and we want to do
+        # math with the numbers here, so just convert to a float.
+        data = super().get_data()
+        return [[item[0], float(item[1])] if len(item) > 1 else item
+                for item in data]

### CODE REVIEW

**Comment by Mr0grog** [comment]
@rickpr a) will need to incorporate this in #151, but b) maybe worth thinking about whether there might be a better, more generic way to do this, e.g. having some property on `PowerBiQuerier` that says “cast each row to these types,” e.g. `[int, float]` in this case.

**Comment by benghancock** [comment]
@Mr0grog I ran this locally and looks good! 

Just so I understand correctly, it looks like we are not capturing the positive percentage figure but instead are calculating the number of positive tests based on that value. Is that right? I haven't looked closely at the Alameda BI dashboard so I'm just inferring based on this and the output:

``` python
# time_series_tests.py
    def _positive_and_negative_tests(self, total_tests: int, percent_positive_tests: int) -> Dict[str, int]:
        positive_tests = round(total_tests * percent_positive_tests / 100)
        negative_tests = total_tests - positive_tests
        return { 'positive': positive_tests, 'negative': negative_tests }
```

Mr0grog:
> it looks like we are not capturing the positive percentage figure but instead are calculating the number of positive tests based on that value. Is that right?

Yep, that’s correct. (I am also just looking at the output and code, but it’s clear that’s what we’re doing. @rickpr originally wrote this code, so he may or may not have more context to add.)

Mr0grog:
Hmmmm, digging in more, it looks like we *can* get a count of positive tests from the API rather than calculating it from %-positives, so that would be more accurate. HOWEVER, it only seems to work for the `V_Tests_RollingSevenDayPercentagePositive` table instead of the `V_Combined_Tests` table. (Looks like you can only look up single values on that one, despite the fact that it does have a date column named `DtCount`. I may just be querying wrong, though. 🤷) On the other hand, it looks like we are doing the math using `V_Tests_RollingSevenDayPercentagePositive` right now anyway, so that’s no *loss* in precision.

Anyway, I think switching that up is a bit much for a hotfix, but maybe it’s a change we should make…

PR #161, opened 2020-12-18
HOTFIX: handle repeat rows in Solano age groups

Mr0grog:
The last day of Solano's data seems to be getting repeated in their dataset (every row is repeated once, totally verbatim, as far as I can tell). I'm not sure why that's happening (it's there that way in the source data if you go look on the web), but it seems benign, so I've just added a fix here to de-duplicate the age groups (we just take the last value we have for each group). This already happens naturally for race/ethnicity and gender (since we build dicts instead of lists), so there's no need to fix those.

### DIFF
--- a/covid19_sfbayarea/data/solano.py
+++ b/covid19_sfbayarea/data/solano.py
@@ -220,21 +220,21 @@ def get_age_table(out: Dict) -> None:
     response2.raise_for_status()
     parsed2 = response2.json()
 
-    # Format the results.
-    age_group_cases = [{
-                        'group': entry['attributes']['Age_group'],
-                        'raw_count': entry['attributes']['AG_Total_cases']
-                       }
-                       for entry in parsed2['features']]
-    age_group_deaths = [{
-                         'group': entry['attributes']['Age_group'],
-                         'raw_count': entry['attributes']['AG_deaths']
-                        }
-                        for entry in parsed2['features']]
+    # Because of the way data is structured (ages, race/ethnicity groups, etc.
+    # all on one line), some categories get repeated (even though the data is
+    # not different). To handle this, create a dict with our groupings to
+    # de-duplicate before formatting final results.
+    age_groups = {entry['attributes']['Age_group']: entry['attributes']
+                  for entry in parsed2['features']}
+
+    age_group_cases = [{'group': group, 'raw_count': data['AG_Total_cases']}
+                       for group, data in age_groups.items()]
+    age_group_deaths = [{'group': group, 'raw_count': data['AG_deaths']}
+                        for group, data in age_groups.items()]
 
     # Validate that we’ve got the right age groups.
     if len(age_group_cases) != 4:
-        raise FormatError(f'Race_eth query did not return 5 groups. Results: {age_group_cases}')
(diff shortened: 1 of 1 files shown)


covid19_sfbayarea/data/solano.py
```diff
@@ -220,21 +220,21 @@ def get_age_table(out: Dict) -> None:
...
-                         'group': entry['attributes']['Age_group'],
-                         'raw_count': entry['attributes']['AG_deaths']
-                        }
-                        for entry in parsed2['features']]
+    # Because of the way data is structured (ages, race/ethnicity groups, etc.
+    # all on one line), some categories get repeated (even though the data is
+    # not different). To handle this, create a dict with our groupings to
+    # de-duplicate before formatting final results.
+    age_groups = {entry['attributes']['Age_group']: entry['attributes']
+                  for entry in parsed2['features']}
```
benghancock, line 228: Just making sure I'm understanding the logic correctly: this will loop over each dict in `features`, and reassign the value for a given key (`Age_group`) if a duplicate occurs, right?

benghancock (commented):
Hey @Mr0grog ! It looks like this fixes the duplicate issue. Nice! 

Alas, we share the same fate as Sisyphus: I'm seeing another error for Solano that seems unrelated; maybe that can be for another PR?

```
Solano county failed: This dashboard has changed. None of the <div> elements contains'Disclaimers' https://doitgis.maps.arcgis.com/apps/opsdashboard/index.html#/d28335cd317a45cd84211cd290889c27
  File "scraper_data.py", line 30, in main
    out[county] = data_scrapers.scrapers[county].get_county()
  File "/home/benh/projects/opensource/data-covid19-sfbayarea/covid19_sfbayarea/data/solano.py", line 33, in get_county
    out["meta_from_source"] = get_notes()
  File "/home/benh/projects/opensource/data-covid19-sfbayarea/covid19_sfbayarea/data/solano.py", line 144, in get_notes
    raise FormatError(
covid19_sfbayarea.errors.FormatError: This dashboard has changed. None of the <div> elements contains'Disclaimers' https://doitgis.maps.arcgis.com/apps/opsdashboard/index.html#/d28335cd317a45cd84211cd290889c27
```

covid19_sfbayarea/data/solano.py
Mr0grog: Yeah. Comprehensions can sometimes be tough to read, so I try not to use them if they get too complex. (Have I done so here?) This is effectively equivalent to:

```py
age_groups = {}
for entry in parsed2['features']:
    key = entry['attributes']['Age_group']
    value = entry['attributes']
    age_groups[key] = value
```

So if you have multiple dicts with the same age group in `features`, the last one will win, since we are making the age group into the key for a new dictionary and then just writing over the same key again and again.

Mr0grog:
Rebased on master.

@benghancock it looks like it’s working for me again now. How about you?

covid19_sfbayarea/data/solano.py
benghancock: Got it! It's clear -- just wanted to make sure I was reading it right!

benghancock:
Yep, working now! Good to go.

PR #167, opened 2021-01-08
HOTFIX: Rewrite Marin data scraper to use Socrata

### LINKED ISSUE #162: Marin county does not include unknown race/age/gender data
The numbers we have for race/ethnicity, age, and gender in Marin are pulled from the charts on [Marin’s dashboard](https://coronavirus.marinhhs.org/surveillance), which do not include unknown race/ethnicities, ages, and genders. For example, race/ethnicities today report as:

| Race/Ethnicity                   | Cases | Case Percent |
| -------------------------------- | ----: | -----------: |
| Hispanic/Latino                  | 3,046 |        59.2% |
| White                            | 1,561 |        30.3% |
| Asian                            |   139 |         2.7% |
| Black/African American           |    91 |         1.8% |
| Native Hawaiian/Pacific Islander |    15 |         0.3% |
| American Indian/Alaska Native    |     7 |         0.1% |
| Multi or Other Race              |   285 |         5.5% |

But if unknowns were included, this would be:

| Race/Ethnicity                   | Cases | Case Percent | Up to |     |
| -------------------------------- | ----: | -----------: | ----: | --- |
| Hispanic/Latino                  | 3,046 |        45.7% | 68.5% | █████████░░░░░ |
| White                            | 1,561 |        23.4% | 46.2% | █████░░░░ |
| Asian                            |   139 |         2.1% | 24.9% | ░░░░░ |
| Black/African American           |    91 |         1.4% | 24.1% | ░░░░░ |
| Native Hawaiian/Pacific Islander |    15 |         0.2% | 23.0% | ░░░░░ |
| American Indian/Alaska Native    |     7 |         0.1% | 22.9% | ░░░░ |
| Multi or Other Race              |   285 |         4.3% | 27.0% | █░░░░░ |
| Unknown                          |  1516 |        22.8% |     - | ████ |

(Where “up to” is the represents the top end of possible values if we knew all the unknowns; e.g. “hispanic/latino” could be anywhere from 45.7% – 68.5%.)

I calculated the above by looking at the total case counts and comparing them to the total among the race/ethnicity counts. We can do the same for ages and genders.

We can also do this for hospitalizations and deaths, but that one gets a little sticky. For cases, hospitalizations, and deaths, we have a point-in-time, “current” count at the top of the dashboard, and we have time series charts lower down. However, those two don’t match up when it comes to hospitalizations and deaths (they do for cases). They are described the same way in both places, so I’m not clear on what the discrepancy is. We need to choose which to go with.

### LINKED ISSUE #165: Switch Marin to Use New Data Portal
We currently [get data for Marin county by scraping their dashboard using Selenium](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/master/covid19_sfbayarea/data/marin.py), but @1ec5 pointed out that Marin has recently set up a Socrata-powered dashboard with COVID data found at: https://data.marincounty.org/browse?q=covid

This ought to be a *much* better and easier to use source of data, so we should rewrite the Marin scraper to use Socrata instead. This *may* also obviate #162 (I haven’t dug in to see how unknown demographic categories are or are not accounted for in the data portal).

If you are new to the codebase and are implementing this, you might want to check out the implementation for two other counties that are Socrata-based:

- [San Francisco](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/master/covid19_sfbayarea/data/san_francisco.py)
- [Santa Clara](https://github.com/sfbrigade/data-covid19-sfbayarea/pull/164/files) (not yet merged)

Mr0grog:
Marin now publishes COVID data in an actual data portal with a real API and changed the dashboard to use Tableau. That broke our scraper (because all the charts are now built in an entirely different way). The best way to fix was to rewrite on top of the Socrata data portal API, which we should happily expect to be much more stable.

This also adds caching to the Socrata API so we can make multiple calls to the same URL without actually making multiple HTTP requests. The Marin data is arranged such that a lot of different dimensions of data are combined (unlike most other portals where they are separated), and caching lets us keep the logic straightforward without making unnecessary repeated requests.

This also turns out to fix #162 -- unknowns are now included in the data!

Fixes #165.

### DIFF
--- a/covid19_sfbayarea/data/marin.py
+++ b/covid19_sfbayarea/data/marin.py
@@ -1,287 +1,278 @@
-#!/usr/bin/env python3
-import csv
-from typing import Any, List, Dict, Generator, Iterable, Tuple
-from urllib.parse import unquote_plus
-from datetime import datetime, timezone
-from contextlib import contextmanager
-from selenium import webdriver # type: ignore
-
+from typing import List, Dict, Iterable
+from datetime import datetime
 from ..errors import FormatError
-from ..utils import PERMISSIVE_SPACES
-from ..webdriver import get_firefox
-from .utils import get_data_model
+from ..utils import assert_equal_sets, parse_datetime
+from .utils import SocrataApi
 
 
-class MarinDashboardPage:
-    """
-    Represents a Selenium Webdriver browser opened to the Marin County COVID-19
-    Dashboard page.
-    """
-    url = 'https://coronavirus.marinhhs.org/surveillance'
-    driver: webdriver.Remote
+API_IDS = {
+    # This timeseries includes cases, deaths, and hospitalizations (not tests)
+    'cases': 'wg8s-i3c7',
+    'tests': 'kr8c-izzb',
 
-    def __init__(self) -> None:
-        self.driver = get_firefox()
-        self.driver.implicitly_wait(30)
-        self._load()
+    # Cumulative demographics for age, race/ethnicity, and gender across cases,
+    # hospitalizations, and deaths are all published in one dataset.
+    # Breakdowns by transmission vector and comorbidities are not available.
+    # NOTE: Cases (but not hospitalizations or deaths) are also available as a
+    # separate timeseries for each demographic breakdown, but we do not
+    # use those at the moment. Most other counties do not publish demographic
+    # timeseries.
+    'demographics': 'uu8g-ckxh'
 
-    def find(self, selector: str) -> Any:
-        "Find the first element matching a CSS selector."
-        return self.driver.find_element_by_css_selector(selector)
-
-    def find_all(self, selector: str) -> List:
-        "Find all element matching a CSS selector."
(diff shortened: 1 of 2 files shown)


benghancock:
This looks really good @Mr0grog! One question I have is whether there's a value in pulling down the "Total Hospitalized" stats. I haven't cross-checked with the state data that the hospital scraper pulls; perhaps we're already covered, although Marin has demographic breakdowns. Just asking because the data is there.

Otherwise, I think it's good to go!

Mr0grog:
@benghancock good point! I think I avoided it because we don’t have any standard around it and adding another timeseries is a lot of extra data which could be an issue for the front-end project. It does seem kinda disappointing to leave out when we have it, though.

Since this is a hotfix, I’m going to take your comment as an approval and merge as-is so we’ve got something working, but I’ll file an issue to add hospitalization. You or anyone else can implement.

Mr0grog:
Filed the issue as #170.

Issue #168, opened 2021-01-09
Santa Clara news scraper is missing news from 2020


Mr0grog:
The Santa Clara news scraper is only showing items from the current year (2021) even if items from the previous year are still within the expected time range (31 days by default). See this diff from a run of the scraper for an example — a huge number of items are being deleted that shouldn’t be: https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/828/files

This is happening because the news page (https://www.sccgov.org/sites/phd/news/Pages/newsroom.aspx) only shows news from a given year (e.g. 2021), and the current scraper simply reads what’s actively shown on the page. There are two obvious fixes I can think of here:

1. Change [`news.santa_clara.SantaClaraNews`](https://github.com/sfbrigade/data-covid19-sfbayarea/blob/6c4585be592f14ed49a85df1557d4568e79188de/covid19_sfbayarea/news/santa_clara.py#L12) to iterate through selecting each value in the “Please select a year” dropdown and scraping the page content. Ideally this should check `self.from_date` and `self.to_date` to ensure it only bothers with looking at years that are in the expected timeframe.

2. Instead of scraping news items from the DOM, read the content from the `<script>` element that starts with `var sccgovAlertsArchiveData = ` (all the news is stored in this script and the DOM is built dynamically). This could be a bit brittle, though: you might be able to pull out the right part of the script and parse it as JSON, but it may be prone to change in ways that make it dramatically harder to parse.

**To Reproduce**

1. Run `python scraper_news.py santa_clara --from 31` (or set `--from` so a number of days that includes December 31, 2020)
2. Check whether output includes a news item from December 31, 2020. It’s buggy if the news item is not shown.

PR #169, opened 2021-01-11
Scrape Santa Clara news from multiple calendar years

slim-builder:
Fix for issue #168


Mr0grog (commented):
Wow, I totally didn’t think just concatenating the page sources together would work (I figured multiple `<html>` or `<body>` elements might gum up the works). This is nice and simple!

Could you please fix up the linting/type errors? Otherwise this is fantastic, thank you, @slim-builder!

Mr0grog (approved):
Looks fantastic! Thank you. 👍 👍

PR #176, opened 2021-01-25
HOTFIX: Napa all news page is broken

Mr0grog:
Napa county's page for all news is broken (`https://www.countyofnapa.org/CivicAlerts.aspx?sort=date`) and doesn't load, so we can't actually scrape it. However, if we look at news from only a subset of categories, it works fine (it appears that category 17 [Board of Supervisors] is broken). This lists all the categories except one that breaks, so the page loads and gives us most news.

Test this by running:

```sh
$ python scraper_news.py napa
```

And making sure it outputs some news instead of an error.

### DIFF
--- a/covid19_sfbayarea/news/napa.py
+++ b/covid19_sfbayarea/news/napa.py
@@ -44,10 +44,14 @@ class NapaNews(NewsScraper):
 
     FEED_INFO = dict(
         title='Napa County COVID-19 News',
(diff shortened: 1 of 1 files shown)


rickpr (approved):
LGTM - I was able to see the error page and the page seems to work fine with the categories limited.

Mr0grog:
🎉 Thanks for reviewing!

PR #180, opened 2021-01-31
Add Napa County data scraper

### LINKED ISSUE #24: New Data: Napa County
County or Counties: Napa County
Data source: https://legacy.livestories.com/s/v2/coronavirus-report-for-napa-county-ca/9065d62d-f5a6-445f-b2a9-b7cf30b846dd/
Output to: ? 
Update frequency/time: Updated at 10:30 AM, Monday through Saturday.

Mr0grog:
This adds data support for Napa County (wow, that was a long time coming). It's mostly based on the ArcGIS feature server identified by @1ec5, which lists *every case* in Napa County.

There are a few questionable bits I’d love feedback on here:

1. All cases have a date for when the test results were received, but a few do not have a date for when the test was *taken*. The county works around this by showing cases by test result received date, while @1ec5 worked around it for Wikimedia Commons by doing the same thing I’m doing here: using the test taken date where available, and the test received date otherwise. I think this makes the most sense, but it will make our numbers differ slightly from the county dashboard’s display.
    - See the Wikimedia Commons data at: https://commons.wikimedia.org/wiki/COVID-19_pandemic_in_the_San_Francisco_Bay_Area#Napa_County
    - And the code: https://github.com/1ec5/external-data-bot/blob/master/scripts/covid19/US/CA/napa-county.sh

2. Napa only publishes test information on a weekly basis. I’ve put that into our standard format by attributing the week’s test to the final day of the week (Saturdays).

3. Similarly, Napa publishes test positivity *rates*, not counts. To make this comparable with other counties, I’ve done the calculation to get counts, and included the rate as a `rate` property on each timeseries entry.

Fixes #24.

### DIFF
--- a/covid19_sfbayarea/data/__init__.py
+++ b/covid19_sfbayarea/data/__init__.py
@@ -2,6 +2,7 @@ from typing import Dict, Any
 from . import alameda
 from . import san_francisco
 from . import marin
+from . import napa
 from . import sonoma
 from . import solano
 from . import san_mateo
@@ -11,7 +12,7 @@ scrapers: Dict[str, Any] = {
     'alameda': alameda,
     # 'contra_costa': None,
     'marin': marin,
-    # 'napa': None,
+    'napa': napa,
     'san_francisco': san_francisco,
     'san_mateo': san_mateo,
     'santa_clara': santa_clara,
--- a/covid19_sfbayarea/data/arcgis.py
+++ b/covid19_sfbayarea/data/arcgis.py
@@ -1,0 +1,105 @@
+import requests
+from typing import Any, Dict, Generator
+from urllib.parse import urljoin
+from .errors import BadRequest
+
+
+class ArcGisFeatureServer:
+    """
+    A thin wrapper around ArcGIS's FeatureServer API. The capabilities are
+    pretty limited, but it gives us a useful wrapper for the data-focused parts
+    of the API that we need.
+
+    Parameters
+    ----------
+    base_url : str
+        The base URL of the server, including the "Unique Service ID". Example:
+        ``'https://services1.arcgis.com/Ko5rxt00spOfjMqj'``
+    """
+    def __init__(self, base_url: str):
+        if not base_url.endswith('/'):
+            base_url += '/'
+        self.base_url = base_url
+        self.service_base_url = urljoin(base_url, 'ArcGIS/rest/services/')
+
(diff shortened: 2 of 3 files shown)


benghancock:
Hi @Mr0grog! This looks really good. Some feedback below:

> The county works around this by showing cases by test result received date, while @1ec5 worked around it for Wikimedia Commons by doing the same thing I’m doing here: using the test taken date where available, and the test received date otherwise.

I think this approach makes sense. The principle of using the most accurate data where possible, and then falling back if need be, is good one. I would also reckon that using the specimen collection date provides a closer picture of how infections were actually unfolding. 

One oddball date I spotted in the data that I thought I'd flag (but doesn't appear related to this mixing):

```javascript
      "cases": [
        {
          "date": "1992-01-05",
          "cases": 1,
          "cumul_cases": 1
        },
```
Not exactly sure what to make of it, but my guess is it was just entered incorrectly. I see the timestamp in the underlying data:

``` javascript
      "attributes" : {
        "DtLabCollect" : 694598400000, 
        "count" : 1
      }
```

> Napa only publishes test information on a weekly basis. I’ve put that into our standard format by attributing the week’s test to the final day of the week (Saturdays).

Sounds like a logical approach.

> Similarly, Napa publishes test positivity rates, not counts. To make this comparable with other counties, I’ve done the calculation to get counts, and included the rate as a rate property on each timeseries entry.

This looks good as well and obviously having the counts will make the testing data easier to visualize.

Think it's all good to go! Extra kudos on the `ArcGisFeatureServer` wrapper. That's really neat!

benghancock (approved):

Mr0grog:
Ohhhh, good catch. Looking at that row, it seems like the report date is good. I’m going to try querying for only collection dates >= 2019-12-01 and in the report date group, including collection dates < 2019-12-01.

1ec5:
> One oddball date I spotted in the data that I thought I'd flag (but doesn't appear related to this mixing):

Oh yeah, I noticed that too. I assume 2021 was miscoded as 1992 somehow, because the 1992-01-05 date cropped up on 2021-01-12. In 1ec5/external-data-bot@0062bd165d88109cdab36d879578f21a9669ceb3, I worked around it by replacing 1992 with 2021 and merging the cases into the 2021 date. There have been numerous other miscoded cases in Alameda and Santa Clara Counties. Sometimes we can infer the intended date, but other times all we can do is drop the data.

PR #182, opened 2021-02-18
Add data scraper for Contra Costa County

### LINKED ISSUE #21: New Data: Contra Costa County
County or Counties: Contra Costa County
Data source: https://www.coronavirus.cchealth.org/dashboard
Output to: ?
Update frequency/time:  Not clear about the frequency. Dashboard says ```Data reported to CCHS as of 4/13/2020 at 11:30 a.m.```

Mr0grog:
There are quite a few complexities and caveats with this one:

- Positive test counts are estimated based on rate (since the county only publishes the rate), similar to Santa Clara.

- Daily test counts, total test counts, and test positivity rates are each separate charts that cover different date ranges! Putting them together as a single series is a bit messy. Worse, daily test counts and totals don't completely agree (the total tests series starts earlier, and roughly the first two weeks of daily tests have `0` for the value, while the difference between total tests on those days is often > 0). After March 18, 2020 they mostly agree, so daily total test counts seems like the better source. I've calculated daily tests based on the difference between days in total tests, rather than using the daily counts directly.

- Contra Costa splits out hispanic ethnicity from the other racial categories, unlike all the other counties, which provide them as a combined cross section. I'm not sure there's a good way to make our own combined version, so I added an new grouping for `ethnicity`, and left the other races in `race_eth` as usual. It's not great for comparing to other counties. (Maybe we could assume "hispanic" is a chunk out of "white" and calculate the cross section that way, but that seems pretty iffy. I'm open to feedback or suggestions.)

- Case race and case ethnicity are published as percentages, unlike the other case demographics and the corresponding death demographics, so I've calculated them based on the totals. Kind of a weird exception (especially since the same thing isn't done for deaths).

Fixes #21.

### DIFF
--- a/covid19_sfbayarea/data/contra_costa.py
+++ b/covid19_sfbayarea/data/contra_costa.py
@@ -1,0 +1,414 @@
+"""
+Contra Costa County
+
+Contra Costa builds its dashboard with Qlik. Qlik can be a bit complicated, so
+we've also got an API client called ``qlik.QlikClient``.
+
+There's not an obvious way to pull directly from whatever raw data the Qlik
+charts are based on, so instead we have to pull the data that backs each chart.
+(For example, if you dig around enough, it appears there is a single table with
+all timeseries, but there's no clear way to get the data from the table.
+Instead, we get the data behind each of the cases by date chart, deaths by date
+chart, and so on.)
+
+How do we determine the chart IDs to pull data from? When browsing one of the
+dashboard pages, you'll a JS file loaded that is named for the dashboard.
+For example, on the https://www.coronavirus.cchealth.org/overview dashboard, a
+JS file named dashboard.cchealth.org/extensions/COVIDDashboard/Overview.js.
+That JS file will have a series of calls to
+``getObject('<dom_id>', '<chart_id>', ...)`` like so:
+
+    overviewApp.getObject('QV01','gumQeX',{noSelections: true, noInteraction: false});
+    overviewApp.getObject('QV02','LuVGBZ',{noSelections: true, noInteraction: false});
+    overviewApp.getObject('QV03','jWjFxe',{noSelections: true, noInteraction: false});
+    overviewApp.getObject('QV04','cWjnGdK',{noSelections: true, noInteraction: false});
+    overviewApp.getObject('QV05','bZFxmu',{noSelections: true, noInteraction: false});
+
+For each line, the first string is the ID of a DOM element on the page that
+contains the chart and the second string is the chart ID in Qlik.
+
+To get the Qlik chart ID for any chart on the page, use your browser's web
+inspector to find the ``<div>`` that holds the chart and get its ID. Example:
+
+    <div class="chart qvobject" id="QV05">...</div>
+
+Then look in the JS file for the line like above that maps that DOM ID to a
+Qlik chart ID. In the example above, the DOM ID was ``'QV05'``, and you can see
+in the example JS that the matching Qlik chart ID would be ``'bZFxmu'``.
+
+Dashboards:
+Overview - https://www.coronavirus.cchealth.org/overview
+           https://dashboard.cchealth.org/extensions/COVIDDashboard/Overview.html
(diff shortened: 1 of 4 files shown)


Mr0grog:
Rebased on master for a cleaner merge.

covid19_sfbayarea/data/contra_costa.py
```diff
@@ -0,0 +1,415 @@
...
+Then look in the JS file for the line like above that maps that DOM ID to a
+Qlik chart ID. In the example above, the DOM ID was ``'QV05'``, and you can see
+in the example JS that the matching Qlik chart ID would be ``'bZFxmu'``.
+
+Dashboards:
+Overview - https://www.coronavirus.cchealth.org/overview
+           https://dashboard.cchealth.org/extensions/COVIDDashboard/Overview.html
+           ID: b7d7f869-fb91-4950-9262-0b89473ceed6
+
+Vavvines - https://www.coronavirus.cchealth.org/vaccine
```
benghancock, line 44:
```suggestion
Vaccines - https://www.coronavirus.cchealth.org/vaccine
```
Typo

covid19_sfbayarea/data/contra_costa.py
```diff
@@ -0,0 +1,415 @@
...
+    })
+
+    total_tests_data = get_chart_data(api, CHART_IDS['total_tests'])
+    for x, y in total_tests_data:
+        day = api.parse_date(x['qNum'])
+        tests[day]['cumul_tests'] = y['qNum']
+
+    # There's no chart with total positive tests, so we use the *rate* to
+    # calculate it. Note also that this is a rolling average, not the actual
+    # rate on the day, so this is not the most accurate. :(
```
benghancock, line 218: Well noted. How should we communicate these types of limitations on the front end?

covid19_sfbayarea/data/contra_costa.py
```diff
@@ -0,0 +1,415 @@
...
+    # Gender and age groups are published as counts, but race and ethnicity are
+    # published as percentages, so we need thet total to calculate counts.
+    gender = get_cases_by_gender(api)
+    total = sum(gender.values())
+
+    return {
+        'gender': gender,
+        'age_group': get_cases_by_age(api),
+        'race_eth': get_cases_by_race(api, total),
+        'ethnicity': get_cases_by_ethnicity(api, total),
```
benghancock, line 281: I think the way you've handled this is sensible, given how the data is presented. It would be nice to know how the categories in the "Ethnicity" and "Race" charts are associated, but my guess is that the county just treats these as separate fields.

covid19_sfbayarea/data/qlik.py
```diff
@@ -0,0 +1,196 @@
...
+                 _type: Optional[Type[BaseException]],
+                 _value: Optional[BaseException],
+                 _traceback: Optional[TracebackType]) -> None:
+        self.close()
+
+    @staticmethod
+    def parse_date(value: int) -> date:
+        """
+        Parse a date from Qlik. Qlik uses MS Excel's date format
+        (days since 1900-01-01 + 2).
```
benghancock, line 189: Nicely spotted! What a wonky format.

benghancock (approved):
This all looks really good @Mr0grog. I think the way you've navigated the limitations and quirks in the data makes sense, and the Qlik API wrapper is pretty slick! A few comments inline but I think it's good to go! 👍

covid19_sfbayarea/data/contra_costa.py
Mr0grog: I tried to note this in the `meta_from_baypd` field, and I think for now that’s the best technical method we have. But it definitely seems like we might be able to structure this stuff better for display (maybe a `notes` field in each section of the data or something?) or could just hard-code similar information in the UI. Not sure. :\

covid19_sfbayarea/data/contra_costa.py
Mr0grog: Yeah, I’m pretty sure these are totally unrelated, as far as the county is concerned. 😞

PR #185, opened 2021-03-09
HOTFIX: San Mateo tests table was renamed

Mr0grog:
It looks like someone renamed the table and columns for test data in San Mateo, which broke the scraper. I've updated with the new names and verified that the data looks at least *roughly* like it used to. (It seems the timeseries now starts a month later than it used to, though. I tried using a `Where` clause to extend it, but no dice. This is just all that's there now.)

This fixes the broken scrapes we are currently experiencing for San Mateo, e.g. https://sfbrigade.slack.com/archives/C01AV92EM47/p1615198414000100


Mr0grog:
Going ahead and merging since this is a hot fix and nobody is reviewing.

PR #196, opened 2021-03-25
Autocorrect dates from 1920 or 1921 to 2020 or 2021

### LINKED ISSUE #194: Santa Clara data is failing with dates from 1920/1921
The Santa Clara County data scraper is currently failing because the county’s dataset has a date from 1921:

```
Santa Clara county failed: Unknown date format: "1921-02-10T00:00:00.000"
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/scraper_data.py", line 30, in main
   out[county] = data_scrapers.scrapers[county].get_county()
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/data/santa_clara.py", line 64, in get_county
   'deaths': get_timeseries_deaths(api),
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/data/santa_clara.py", line 95, in get_timeseries_deaths
   return [
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/data/santa_clara.py", line 97, in <listcomp>
   'date': parse_datetime(entry['date']).date().isoformat(),
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/utils.py", line 48, in parse_datetime
   raise ValueError(f'Unknown date format: "{date_string}"')
ValueError: Unknown date format: "1921-02-10T00:00:00.000"
```

We’ve seen similar issues before with dates from 1920 as well. They usually get resolved by the county, but it can take several days or more for that to happen, and it’s stupid for the scraper to be broken for that whole time.

I think it’s relatively safe to assume someone accidentally entered 2-digit year in some form where a 4-digit one was required, so we can safely change these to *2020* or *2021*.

Mr0grog:
We often see dates from 1920 and 1921 that seem to be someone accidentally typing two digits into a four-digit date field, and the system storing it as a century earlier. Since this seems like a common issue, we can probably just autocorrect it to be a century later.

Fixes #194.

### DIFF
--- a/covid19_sfbayarea/utils.py
+++ b/covid19_sfbayarea/utils.py
@@ -28,7 +28,12 @@ def dig(items: Union[Dict[Any, Any], List[Any]], json_path: List[Any]) -> Any:
         raise KeyError(f'Error reading data at path: {json_path}') from error
 
 
-def parse_datetime(date_string: str, timezone: Optional[tzinfo] = PACIFIC_TIME) -> datetime:
+def parse_datetime(
+    date_string: str,
+    timezone: Optional[tzinfo] = PACIFIC_TIME,
+    *,
+    correct_century: bool = True
+) -> datetime:
     """
     Parse a datetime from a string and ensure it always has a timezone set. Use
     the `timezone` argument to set the timezone to use if none was specified in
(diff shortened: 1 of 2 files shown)


covid19_sfbayarea/utils.py
```diff
@@ -28,7 +28,12 @@ def dig(items: Union[Dict[Any, Any], List[Any]], json_path: List[Any]) -> Any:
         raise KeyError(f'Error reading data at path: {json_path}') from error
 
 
-def parse_datetime(date_string: str, timezone: Optional[tzinfo] = PACIFIC_TIME) -> datetime:
+def parse_datetime(
+    date_string: str,
+    timezone: Optional[tzinfo] = PACIFIC_TIME,
+    *,
+    correct_century: bool = True
```
rickpr, line 35: Do we need this argument or should we always correct it?

rickpr (approved):
Doesn't seem like there was a lot of relevant data in the 20th century!

covid19_sfbayarea/utils.py
Mr0grog: I *think* we would always correct, but I wanted it to be there in case we ever needed to deal with some historical data or something, so it would be easy to turn off.

PR #199, opened 2021-04-06
HOTFIX: Get San Mateo deaths by day from LA Times

### LINKED ISSUE #190: Get San Mateo deaths timeseries from LATimes or NYTimes
In #189, I hotfixed San Mateo to get its deaths timeseries from CHHS (the county does not publish timeseries data for deaths). However, the old state data matched the county’s data, while CHHS’s data is *really* far off (342 vs. 534, or off by over 40%!). In contrast, both the LATimes and NYTimes publish a timeseries that matches up with the county. We should probably switch to one of them. (I’d lean towards LATimes since they are more CA-focused.)

- LATimes data: https://github.com/datadesk/california-coronavirus-data/blob/master/latimes-county-totals.csv
- NYTimes data: https://github.com/nytimes/covid-19-data/blob/master/us-counties.csv

Mr0grog:
The state has ceased publishing COVID-19 deaths as a timeseries, so we need to load the data from another source. LA Times's data appears to have been more consistent with the county than the data we were previously loading from CHHS, so this not only is a hotfix, but probably improves our data quality anyway.

Fixes #190.

### DIFF
--- a/covid19_sfbayarea/data/san_mateo/__init__.py
+++ b/covid19_sfbayarea/data/san_mateo/__init__.py
@@ -2,7 +2,7 @@ import csv
 import json
 import requests
 
-from datetime import datetime, date
+from datetime import datetime
 from typing import Any, Dict, List, cast
 from covid19_sfbayarea.utils import dig, parse_datetime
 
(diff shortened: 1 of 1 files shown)


rickpr (approved):
Tested out locally and works great!

PR #208, opened 2021-05-06
HOTFIX: Solano is OK if disclaimers are missing

Mr0grog:
The Solano data scraper looks for disclaimer text on the dashboard, and fails if there is none. Sometime in the last few days, they appear to have removed all the disclaimers. The data was actually coming through fine otherwise.

I switched this to log a warning, but maybe this case isn’t really exceptional enough to even do that. Either way, I figure this should be fine for a hotfix.


benghancock (approved):
Looks good! Nice quick fix 🔧

PR #210, opened 2021-05-25
HOTFIX: Alameda news has language link as summary

Mr0grog:
One of the news items on the Alameda news page has a `<br>` tag inside one of the language-specific links for the news item. This is causing the parser to flip back to the summary parsing state and read the text of the link as the news item's summary. This fixes the issue by jumping to the next sibling of the language link after parsing it, rather than unnecessarily moving *inside* the link. (Usually that’s no issue, because the text inside the link doesn't cause any special behavior.)


Mr0grog:
This is addressing the issue seen in this PR: https://github.com/sfbrigade/stop-covid19-sfbayarea/pull/1214/files

PR #215, opened 2021-06-12
HOTFIX: Ignore SSL Cert errors for Contra Costa

Mr0grog:
The Contra Costa scraper is currently failing because of SSL certificate verification problems. Ignoring SSL certificate errors resolves the issue with relatively minimal fuss.


benghancock (approved):
Looks good! Thanks for putting in the fix!

PR #216, opened 2021-06-25
Remove warning about no disclaimers in Solano

Mr0grog:
Right now, we get this message on the #proj-covid-alerts channel every day, and it’s not actually very useful:

> None of the elements on the dashboard contain "disclaimer?" (https://doitgis.maps.arcgis.com/apps/opsdashboard/index.html#/d28335cd317a45cd84211cd290889c27)

It’s coming from Solano County scraper and isn't really signifying a problem, but it logs every night when we scrape and makes it appear as if there *is* a problem. I’ve simply removed the warning.


Mr0grog:
Going ahead and merging this since it’s been sitting for a while, and has pretty minimal impact on functionality.

PR #217, opened 2021-06-25
HOTFIX: Contra Costa no longer has full history

Mr0grog:
The Contra Costa dashboard no longer shows data that covers a complete history of cases, so the first record has a higher cumulative number of cases than just the cases on that day. We have a sanity checking/reconciliation process that checks that, which is now causing the scraper to fail. This fixes the situation by no longer trying to reconcile the daily and cumulative cases on the first day of the dataset. (This also fixes an error we encounter while trying to *report* this failure: we checked the wrong property name for the date.)

This fixes the following error we are currently [seeing in Slack](https://sfbrigade.slack.com/archives/C01AV92EM47/p1624615435000100):

```
Contra Costa county failed: 'day'
Traceback (most recent call last):
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/scraper_data.py", line 30, in main
   out[county] = data_scrapers.scrapers[county].get_county()
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/data/contra_costa.py", line 112, in get_county
   'cases': get_timeseries_cases(api),
 File "/home/runner/work/stop-covid19-sfbayarea/stop-covid19-sfbayarea/scraper/covid19_sfbayarea/data/contra_costa.py", line 157, in get_timeseries_cases
   raise FormatError(f'Sum of daily cases != cumul_cases at record {index} (date: {record["day"]})')
KeyError: 'day'
```

Also see the GitHub actions logs directly: https://github.com/sfbrigade/stop-covid19-sfbayarea/runs/2913050076?check_suite_focus=true

PR #218, opened 2021-07-04
HOTFIX: messy data in Santa Clara and Contra Costa

Mr0grog:
This fixes issues that have cropped up in two counties (Santa Clara and Contra Costa) that are breaking their scrapers:

- In Contra Costa, the specifics of the issue I worked around in #217 have changed slightly -- now the first day with data is prefixed by a month of records with case counts of 0. It seems like these are probably fake, since the first day with data still exhibits the same problem that the first day used to, that its total is clearly accounting for some earlier records that are not shown in the public data we can see. The fix is basically the same, except it's no longer about the *first* day; it's about the first day with non-zero data.

- In Santa Clara, the final row has no date. Since it's only one row and will hopefully be fixed on the county's side, I've changed the code to just skip the record and log a warning, but otherwise operate fine. If the warning would have applied to *every* record, though, we know something else is wrong and it raises a fatal error.

### DIFF
--- a/covid19_sfbayarea/data/contra_costa.py
+++ b/covid19_sfbayarea/data/contra_costa.py
@@ -152,10 +152,16 @@ def get_timeseries_cases(api: QlikClient) -> List[dict]:
     # Sanity-check daily cases vs. totals
     total = 0
     for index, record in enumerate(result):
-        # We can't find a reliable county-level source that covers all time, so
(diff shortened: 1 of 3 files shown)


Mr0grog:
I’m going to go ahead and merge this without a review once tests pass since it’s a hotfix, and it looks like these have been failing for a few days.

